PR #45, opened 2018-01-31
GH-43 download legacy build for unknown os

### LINKED ISSUE #43: Use 'legacy' package as fallback
https://github.com/nodkz/mongodb-memory-server/blob/3fb53263d50d64bb61f621075d9b9420843fe412/src/util/MongoBinaryDownloadUrl.js#L85

it worth to simply return empty string to fall back to 'legacy' distribution instead of throwing an error.

wingedfox:
fixes #43


nodkz:
Thanks! Merged and will be bumped to npm from minute to minute.

PR #49, opened 2018-02-14
Handle newest format of archive name for latest mongodb versions on osx

mathieug:


nodkz:
Thanks for PR!
Please add test cases for 3.6 and 3.4 versions:
https://github.com/nodkz/mongodb-memory-server/blob/master/src/util/__tests__/MongoBinaryDownloadUrl-test.js#L47

mathieug:
Okay I've changed my mind, I'm adding an option "ssl" instead of checking the version. I'm going to push in a few.

nodkz:
Don't forget to add additional option to https://github.com/nodkz/mongodb-memory-server#available-options

Thanks.

mathieug:
Here you go @nodkz. I've also removed the useless binary option http.

mathieug:
I have some troubles with typescript, I need to specify that ssl is optional and default to false.

nodkz:
Thanks a lot! Good work 👌

PR #53, opened 2018-02-25
fix(typo): from "mongoServer" to "mongod"

leonardomerlin:
Sometimes you use the name `mongoServer` and other `mongod` to represent an instance of MongodbMemoryServer.

How could we improve the consistency of documentation?

> PS.: mongod remembers me mongo-daemon :smile:


nodkz:
Thanks!

Issue #55, opened 2018-03-21
MongoDB download not working for new versions


nileshmali:
When I try MongoDB version 3.6.3 on Win 7 64 bit with or without SSL, It did not work.
Seems like dowload URL scheme has changed on server.

rcalabro:
I am getting a 403 .... checked in browser and the url has changed

```
Mongo[54984] Starting MongoDB instance with following options: {"port":54984,"dbName":"test","uri":"mongodb://localhost:54984/test","storageEngine":"ephemeralForTest","dbPath":"/var/folders/ly/5ld4fdyn33d0zb00__75n_q40000gn/T/mongo-mem-43094W4I42INCceiq"} +0ms
Downloading MongoDB: https://downloads.mongodb.org/osx/mongodb-osx-x86_64-3.6.3.tgz
  Mongo[54984] renamed /Users/***/.mongodb-binaries/mongodb-osx-x86_64-3.6.3.tgz.downloading to /Users/***/.mongodb-binaries/mongodb-osx-x86_64-3.6.3.tgz +265ms
(node:43094) UnhandledPromiseRejectionWarning: StatusCodeError: 403 - "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>B7DE450EEECE44FC</RequestId><HostId>CDNqYJFE+Fklv6AXKoaoCDqp2Z63S6d5vzVZx0Da7nmg/GHakCw1O6Pzkn/FeUeJ0JTivGLePsk=</HostId></Error>"
    at new StatusCodeError (/Users/***i/node_modules/request-promise-core/lib/errors.js:32:15)
    at Request.plumbing.callback (/Users/***/node_modules/request-promise-core/lib/plumbing.js:104:33)
    at Request.RP$callback [as _callback] (/Users/***/node_modules/request-promise-core/lib/plumbing.js:46:31)
    at Request.self.callback (/Users/***/node_modules/request/request.js:186:22)
    at Request.emit (events.js:180:13)
    at Request.<anonymous> (/Users/***/node_modules/request/request.js:1163:10)
    at Request.emit (events.js:180:13)
    at IncomingMessage.<anonymous> (/Users/***/node_modules/request/request.js:1085:12)
    at Object.onceWrapper (events.js:272:13)
    at IncomingMessage.emit (events.js:185:15)
    at endReadableNT (_stream_readable.js:1101:12)
[...]

Issue #51, opened 2018-02-23
Download url format may be broken


Koleok:
Looks like the download url format may have changed and would need to be updated.

Recently when running tests I have been seeing the download fail for mongo binary, indeed if i follow the link _(in my case [3.6.0](https://downloads.mongodb.org/osx/mongodb-osx-x86_64-3.6.0.tgz))_ logged for the download I am seeing a standard s3 empty page like:

![image](https://user-images.githubusercontent.com/10928312/36601165-67ba3f2e-1882-11e8-9ed6-31ed45eba274.png)

Log output is as follows using this package with `{ debug: true }`:

```shell
Mongo[58989] Starting MongoDB instance with following options: {"port":58989,"dbName":"01d7602e-c235-4d6b-9009-db9f21c189f0","uri":"mongodb://localhost:58989/01d7602e-c235-4d6b-9009-db9f21c189f0","storageEngine":"ephemeralForTest","dbPath":"/var/folders/q2/5tmxbvwj6gs33d926p9_m45h0000gn/T/mongo-mem-51484gWf8ctpeqK84"} +0ms
  console.warn node_modules/node-uuid/uuid.js:48
    [SECURITY] node-uuid: crypto not usable, falling back to insecure Math.random()

  console.log node_modules/mongodb-memory-server/lib/MongoMemoryServer.js:99
    Autostarting MongoDB instance...

  console.log node_modules/mongodb-memory-server/lib/util/MongoBinaryDownload.js:172
    Downloading MongoDB: https://downloads.mongodb.org/osx/mongodb-osx-x86_64-3.6.0.tgz

  Mongo[58989] renamed ~/.mongodb-binaries/mongodb-osx-x86_64-3.6.0.tgz.downloading to ~/.mongodb-binaries/mongodb-osx-x86_64-3.6.0.tgz +8s
  Mongo[59099] Starting MongoDB instance with following options: {"port":59099,"dbName":"e10683f2-87ed-4816-9320-2843d1c6fe35","uri":"mongodb://localhost:59099/e10683f2-87ed-4816-9320-2843d1c6fe35","storageEngine":"ephemeralForTest","dbPath":"/var/folders/q2/5tmxbvwj6gs33d926p9_m45h0000gn/T/mongo-mem-514841O3FXUa6dBXX"} +0ms
  console.log node_modules/mongodb-memory-server/lib/MongoMemoryServer.js:99
    Autostarting MongoDB instance...

[...]

PR #57, opened 2018-03-28
fix(MongoBinaryDownloadUrl): removed `ssl` from options

YozhikM:
Referenced to this [issue](https://github.com/nodkz/mongodb-memory-server/issues/55)
Now binaries for OSX older than 3.0 version is immediately installed with the SSL. Before and including 3.0 - without.

Since the SSL is not used anywhere else, it is completely removed from the code

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,6 @@ const mongod = new MongodbMemoryServer({
     downloadDir?: string, // by default %HOME/.mongodb-binaries
     platform?: string, // by default os.platform()
     arch?: string, // by default os.arch()
-    ssl?: boolean, // by default false (use it if mongodb dist has only ssl version binary)
     debug?: boolean, // by default false
   },
   debug?: boolean, // by default false
--- a/src/MongoMemoryServer.js
+++ b/src/MongoMemoryServer.js
@@ -22,7 +22,6 @@ export type MongoMemoryServerOptsT = {
     downloadDir?: string,
     platform?: string,
     arch?: string,
-    ssl?: boolean,
     debug?: boolean | Function,
   },
   debug?: boolean,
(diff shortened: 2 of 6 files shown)


nodkz:
Thanks! 👌👍

PS. Crazy man! When do you sleep?!

YozhikM:
@nodkz While you're asleep, your enemy increases the level

mathieug:
Before implementing the ssl option, I started to check the mongodb version like this PR. I think it's better to let the choice to the user.
Maybe you could automatically set the default value but let the user change it.

nodkz:
@mathieug for information, all available links of MongoDB binaries:
Osx https://www.mongodb.org/dl/osx
Linux https://www.mongodb.org/dl/linux
Win https://www.mongodb.org/dl/win32

`ssl` option worked only with OSX and if you check available versions of binaries you may see that only several version between 3.0 and 3.4 have both ssl and no ssl binaries. So this option brings more mess than helps.

Can you provide use-case where `ssl` or `no-ssl` version is important?

mathieug:
I know, I have added the ssl option to this package.

I think some users could want to specify ssl or no-ssl with a specific version and I didn't want to prevent them.
For my use, I don't care I always use the latest version.

nodkz:
> For my use, I don't care I always use the latest version.

Really, for testing purposes (the main aim of this package) it's no matter what type `ssl` or `not-ssl` version of binary have been started. It just should work. We had met with the various type of binary URL problems in #55 #51 and maybe several other issues. So the most comfortable and easy way to fix all problems was reducing the logic of determining correct download url. And this decision affected that we deleted ssl option. 

I think that YAGNI principle is exactly this case ;)

mathieug:
I agree since I use the latest version which is only ssl. I just thought it could be this new behaviour by default with still an option to force it.

PR #70, opened 2018-08-17
fix #69 : change default mongodb version to latest

### LINKED ISSUE #69: MongoDB binaries mismatch checking
in postInstall.js
```javascript
if (isModuleExists('./lib/util/MongoBinary')) {
  const MongoBinary = require('./lib/util/MongoBinary').default;

  console.log('mongodb-memory-server: checking MongoDB binaries cache...');
  MongoBinary.getPath({
    version: 'latest',
  }).then(binPath => {
    console.log(`mongodb-memory-server: binary path is ${binPath}`);
  });
}
```
this will be download mongodb latest version to skip timeout setup `jasmine.DEFAULT_TIMEOUT_INTERVAL = 60000;`

but in /util/MongoBinary.js
```javascript
export default class MongoBinary {
  static cache: MongoBinaryCache = {};

  static async getPath(opts?: MongoBinaryOpts = {}): Promise<string> {
    const {
      downloadDir = path.resolve(os.homedir(), '.mongodb-binaries'),
      platform = os.platform(),
      arch = os.arch(),
      version = '3.4.4',
} = opts;
```
the mongodb default version is 3.4.4

Therefore, it will always download mongodb-binaries version 3.4.4 if not define a version on the first run

If we make this 2 points (mongodb version) to the same version then this problem will solved?

BankyKmitl:


nodkz:
Thanks a lot! For the help!

🎉 This PR is included in version 2.0.0 🎉

Version 1.9.4 was unpublished. Your changes may bring BREAKING CHANGES to someone tests (for most users will no any problems). Also, I drop Node v4 support.

PR #72, opened 2018-08-21
Pass port as object to get-port

### LINKED ISSUE #71: MongoDB always runs on default random port
In memory Mongo is instantiated like this:
`const mongod = new MongodbMemoryServer.default({
    instance: {
        dbName: 'my-test-db',
        port: 12345
    }
})`

It's expected to have MongoDB running on port 12345, but that's not the case. Calling `getConnectionString` and `getPort` show that MongoDB is actually running on some default random port (e.g. `mongodb://localhost:54271/my-test-db`).

This started happening on version `1.9.3` and also happens on `2.0.0`.

Lanchi:
According to `get-port` [docs](https://www.npmjs.com/package/get-port) preferred  port should be passed in as object.

Resolves #71

### DIFF
--- a/src/MongoMemoryServer.js
+++ b/src/MongoMemoryServer.js
@@ -109,7 +109,7 @@ export default class MongoMemoryServer {
     let tmpDir;
 
     const instOpts = this.opts.instance;
-    data.port = await getport(instOpts.port);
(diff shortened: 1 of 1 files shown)


nodkz:
Thanks a lot!
🍻

PR #77, opened 2018-09-13
fix #76: Add reading environment variable on npm postinstall hook

### LINKED ISSUE #76: How I can change binary.downloadDir through env?
https://github.com/nodkz/mongodb-memory-server/blob/bbf75b838f2e3a92a9f4c5e11d805e1c3c2d9c80/src/util/MongoBinary.js#L25-L30

Please define this options as follow

```javascript
const {
  downloadDir = process.env.MONGOMS_DOWNLOAD || path.resolve(os.homedir(), '.mongodb-binaries'),
  platform = process.env.MONGOMS_PLATFORM || os.platform(),
  arch = process.env.MONGOMS_ARCH || os.arch(),
  version = process.env.MONGOMS_VERSION || 'latest',
} = opts;
```
otherwise how I can change it options on `npm install` from CI configurations for `Gitlab`? It can't cache files  from project outside, for caching `/root/.mongodb-binaries` folder. Now I can only change my test files as
```javascript
before(async () => {
  server = new MongoMemoryServer({ binary: { downloadDir: './.mongodb-binaries' } });
});
```
But this test will fail by timeout cause start download binaries. And this does not working because on `postinstall` script binaries anyway will downloads to `os.homedir() + '/.monodb-binaries'`, than on test start binaries will downloads again to `/path/to/project/.mongodb-binaries`. Than Gitlab does not cache this binaries because job is fail.

vdakalov:


nodkz:
Can I ask you to move `ENV_VARS` at the end of `### Available options` section? After all options.
Tnx.

vdakalov:
Tests fails always. It's ok?

src/util/MongoBinary.js
```diff
@@ -22,12 +22,23 @@ export default class MongoBinary {
...
+        ? process.env.MONGOMS_DOWNLOAD_DIR
+        : path.resolve(os.homedir(), '.mongodb-binaries'),
+      platform: process.env.hasOwnProperty('MONGOMS_PLATFORM')
+        ? process.env.MONGOMS_PLATFORM
+        : os.platform(),
+      arch: process.env.hasOwnProperty('MONGOMS_ARCH') ? process.env.MONGOMS_ARCH : os.arch(),
+      version: process.env.hasOwnProperty('MONGOMS_VERSION')
+        ? process.env.MONGOMS_VERSION
+        : 'latest',
+      debug: process.env.hasOwnProperty('MONGOMS_DEBUG')
```
nodkz: For error fix you need to change this line:
```diff
- process.env.hasOwnProperty('MONGOMS_DEBUG')
+ (typeof process.env.MONGOMS_DEBUG === 'string')
```

nodkz: When I merge changes to master Travis will publish a new package. With broken tests - will not. So Travis tests should pass without errors in your PR.

You may run `yarn test` for locally testing on your dev machine.

vdakalov: It's very strange fix! =)
I wonder how it affected to tests?

vdakalov: I will not resolve it fix because it is not valid for eslint. I make explicit commit.

nodkz: Hm, I'll clone your fork and look at problem closely.

Issue #74, opened 2018-08-27
Running as a single node replica set


sebastian-nowak:
Hello,

Is it possible to run `mongodb-memory-server` as a standalone replica set, so that it could be used to test transactions?

I think it should be easy to add: https://docs.mongodb.com/manual/tutorial/convert-standalone-to-replica-set/

Regards,
Sebastian

nodkz:
It's a good feature request. 👍

You need to write a new module `MongoInstancesReplicaSet.js`:
1) Should reuse existed methods in https://github.com/nodkz/mongodb-memory-server/blob/master/src/util/MongoInstance.js 
2) Will be nice if it will have the same Interface like use `MongoInstance`
3) Somehow provide additional properties `prepareCommandArgs` for `MongoInstance`
4) Tune `run()` method for starting and awaiting all replicas to be ready.
5) Tune `kill()` method for killing all mongod processes
6) Will be nice if this module will be covered by tests

Sorry, I have no bandwidth for this feature realisation.
But will be happy if somebody adds it.

nodkz:
But for the fast start, you need to improve `prepareCommandArgs` method https://github.com/nodkz/mongodb-memory-server/blob/master/src/util/MongoInstance.js#L76 It should allow to provide additional options, like `replicaSet` name.

mathieug:
👍 great request.
I'll see if I can help with a PR.

jloveridge:
I have created a pull request that makes it possible to run the `mongod` server in replica set mode. In my implementation does not attempt to address changes to `run()` or `kill()` that was mentioned above. I have augmented `prepareCommandArgs` to handle passing of `auth` to run the server with auth enabled if desired, `args` for passing arbitrary arguments such as `--notablescan`, and `replSet` to allow specifying the replica set name. Tests were added to make sure the above changes to `prepareCommandArgs` are handled correctly.

How I use this in my project:
In my `mongo-environment` file that I use with `jest` I added the following to the environment config that I export:
```
async initReplSet() {
        if (!this.global.__MONGO_REPLSET_NAME) { return; }
        const parsed = muri(this.global.__MONGO_URI__);
        const members = parsed.hosts.reduce((pv, cv, idx) => {
            const member = {_id: idx, host: `${cv.host}:${cv.port}`};
            pv.push(member);
            return pv;
        }, []);
        const rsConfig = {
            _id: 'rsTest',
            members,
        };
[...]

PR #79, opened 2018-09-27
Make it possible to run with auth, as a replica set, and with additional args.

jloveridge:
feat(MongoInstance): add new option `instance.args` to allow passing additional args.
feat(MongoInstance): add new option `instance.auth` to allow running with auth enabled.
feat(MongoInstance): add new option `instance.replSet` to specify replica set name.


nodkz:
Related comment https://github.com/nodkz/mongodb-memory-server/issues/74#issuecomment-424960021

@jloveridge Thanks for PR I ready to merge it!

But if you able to write a new class for ReplicaSet - MongoReplicaSet.js and put it near `MongoMemoryServer.js` with combined code from your comment above and existed code from `MongoMemoryServer` it will be amazing.

Let me know to merge PR or wait for MongoReplicaSet class from you.

jloveridge:
@nodkz I don't necessarily have an issue with creating another class. I do question whether or not it is necessary. The functionality I added is specific to passing command line arguments to the spawned `mongod` process. I don't really feel that justifies a separate class as we are still just spawning a single `mongod` instance. Now, if you are wanting to add some tooling around spinning up a replica set instead of just a single instance that is where I feel an additional class might be applicable. I will outline a brief overview of what something like that might look like.

* New class `MongodbMemoryReplSet`.
    * This server type differs from the standard `MongodbMemoryServer` in the following ways:
        * Internally it will use 1 or more instances of `MongodbMemoryServer`.
        * Starting the server will spin up the instances of `MongodbMemoryServer` and initialize
           the replica set. This would be done by performing the initialization on the first instance created
           and then adding each of the other instances as members. This will necessarily rely on the ability
           to pass some of the very configuration options that I just made possible in these changes for
           `MongodbMemoryServer` and `MongodbInstance `.
        * Starting, stopping, and initialization of the replica set members would all be handled internally.
        * Starting the server would not resolve until all members have been spun up and the replica set
           is in a healthy state.
        * Add convenience methods to allow step-down and simulated failure of replica set members.
        * `getConnectionString` would return a string that includes all of the replica set members and
             look something like: `mongodb://127.0.0.1:27333,127.0.0.1:27334,127.0.0.1:27335/dbName`.
* New options type: `MongoMemoryReplSetOptsT`.
    * Differences from `MongoMemoryServerOptsT`
        * Instead of an `instance` parameter there would be `instances` which would be an
           array of options for each instance that will be spawned.
        * Add a `replSet` property that allows specifying replica set specific configuration items such
           as the name of the replica set (`--replSet`), oplog size (`--oplogSize`), and possibly the number
           of member nodes to spin up (this one has no analog to `mongod` command line params).

So, while I do believe the above additions would be nice I feel they are outside the scope of just being able to pass additional command line arguments when starting an instance of `mongod` which is effectively what this PR addresses. I would suggest we merge this PR as is and a separate issue be created to make spinning up and management of a replica set less onerous on people who want/need them. As I commented on issue #74 it is fairly easy for someone to utilize what I have already done and get up and running with a single node replica set right now.

jloveridge:
I should have mentioned that I would be happy to tackle the creation of the new `MongoMemoryReplicaSet` implementation that I mentioned above. I just feel it should be addressed separately from this PR.

nodkz:
@jloveridge Agreed! Let move new class to another PR. 
Please add info about new options to README.md and I'll publish a new version.

Tnx.

nodkz:
I'm already past midnight. But about 20 minutes will be online. And eeady to merge changes from mobile.

PS. Your overview is brilliant!

jloveridge:
Yep, going to add type definitions now since you have merged my other PR. Just need a few minutes.

jloveridge:
Types have been added. Build passed for me locally. I expect Travis CI will pass very soon.

nodkz:
It merged with master and after CI run it will publish a new version.

Thanks a lot for great PRs!! 👍

Now I can go sleep 😉

jloveridge:
Thank you very much. Now I can use the official package instead of a fork. Sleep well.

StefanoDeVuono:
I noticed this doesn't actually pass the '--auth' option to the command line. Is this intentional? I use it sometimes and I'm working on a PR to do just that, if anyone else is interested.

Issue #28, opened 2017-11-06
TypeScript support


nodkz:
Will be awesome if somebody adds TypeScript declarations for better DX.

Like it did @toverux for graphql-compose:
- https://github.com/nodkz/graphql-compose/issues/81
- https://github.com/nodkz/graphql-compose/pull/82

Thanks!

PR #80, opened 2018-09-27
Native TypeScript support

jloveridge:
I have added typescript definitions directly to the package as requested in #28.


jloveridge:
We use TypeScript a lot here. I am very happy to help get TypeScript types into packages we use so I don't have to bother installing them from DefinitelyTyped. I guess now I need to go make a PR for them to remove the types they have for this project since you have merged the PR.

PR #82, opened 2018-09-30
Emit ‘init’, ‘running’, and ‘stopped’ to allow for easier cleanup of listeners.

jloveridge:
It was apparently way too late, 2 AM, when I submitted my last PR. I cleaned this up to emit state changes as explicit events. This makes it easier to clean up event listeners, and makes it so the `waitUntilRunning` method doesn't add and never remove an event listener. Sorry about the original version being less ideal.


nodkz:
Thanks! 👍

It's a normal situation to forget something or make an error. Main priority to ship new features and error fixes fastly from contributors. Most OSS projects died if not changed or changed very late and have tons of opened huge PRs.

nodkz:
For me better to publish many small versions than try to ship one package per month.

PR #83, opened 2018-10-01
chore(exports): Add MongoMemoryReplSet to package exports.

jloveridge:
Apparently I forgot to copy the export details from the flow version to the TypeScript definitions for the package's index file. Whoops!


nodkz:
😱😂😉

Issue #89, opened 2018-10-14
Does it work with mongodb 4?


YuriGor:
Latest working version of mongo I tried is 3.6.3
When I try latest or 4.0.0/1/2/3 - it downloads binaries and then silently exits the process

(K)Ubuntu 18.04
Node v10.7.0
mongodb-memory-server 2.4.3

AJRdev:
@YuriGor Did you specified your mongodb version with the env variable **MONGOMS_VERSION** ?

YuriGor:
No, I used default constructor parameter like this:
```javascript
var mongod = new MongodbMemoryServer.default({
    instance: {
      dbName: 'slug',
    },
    binary: {
      version: '3.6.3',
    },
  });
```
here is my test script:
https://gist.github.com/YuriGor/cc8114a0d0b1ad151eaf58baab24aa3a
(actually it was created for testing possible bug in mongoose)

AJRdev:
In your test script, didn't you forgot to add a await keyworkd to this line ? 
```javascript
mongoose.connect(
    mongoUri,
    { useNewUrlParser: true },
  );
```

YuriGor:
connect is [synchronous method](https://mongoosejs.com/docs/api.html#mongoose_Mongoose-connect)

It has a callback function parameter to handle errors, and I added one:
https://gist.github.com/YuriGor/cc8114a0d0b1ad151eaf58baab24aa3a#file-pre-update-js-L26
but with no effect, it still works fine with 3.6 3 and stops silently with 4.0.*

I also tried to add 'await' also with no success.

AJRdev:
@YuriGor Where do you see that is a synchronous method ? In the mongoose documentation, it is specified that is a Promise that is returned so it suggests that is a asynchronous operation.
But the docs also specify that if you wdon't wait the connect method to resolve you will be able to define mongoose schemas but the first operation that you will do on the database will hang on until mongoose is not connected to the database and will not throw any error : 
https://mongoosejs.com/docs/connections.html
> Mongoose lets you start using your models immediately, without waiting for mongoose to establish a connection to MongoDB.
> 
> mongoose.connect('mongodb://localhost:27017/myapp');
> var MyModel = mongoose.model('Test', new Schema({ name: String }));
> // Works
> MyModel.findOne(function(error, result) { /* ... */ });
[...]

PR #96, opened 2018-10-25
List latest Ubuntu version

YuriGor:
#89

### DIFF
--- a/src/util/MongoBinaryDownloadUrl.js
+++ b/src/util/MongoBinaryDownloadUrl.js
@@ -200,8 +200,6 @@ export default class MongoBinaryDownloadUrl {
       name += '1604';
     } else if (majorVer === 16) {
       name += '1604';
-    } else if (os.release === '18.04') {
-      name += '1804';
     } else if (majorVer === 18) {
       name += '1804';
     } else {
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ node_modules
 
 coverage
 .nyc_output
+package-lock.json
--- a/src/util/__tests__/MongoBinaryDownloadUrl-test.js
+++ b/src/util/__tests__/MongoBinaryDownloadUrl-test.js
@@ -117,6 +117,14 @@ describe('MongoBinaryDownloadUrl', () => {
         })
       ).toBe('ubuntu1204');
     });
+    it('should return a archive name for Ubuntu 18.04', () => {
+      expect(
+        downloadUrl.getUbuntuVersionString({
+          dist: 'Ubuntu Linux',
+          release: '18.04',
+        })
+      ).toBe('ubuntu1804');
+    });
   });
 
   describe('getDebianVersionString()', () => {


src/util/MongoBinaryDownloadUrl.js
```diff
@@ -200,6 +200,10 @@ export default class MongoBinaryDownloadUrl {
       name += '1604';
     } else if (majorVer === 16) {
       name += '1604';
+    } else if (os.release === '18.04') {
```
nodkz: Please remove this block. 
Check by major version will be enough.

YuriGor: ok

src/util/MongoBinaryDownloadUrl.js
```diff
@@ -200,6 +200,8 @@ export default class MongoBinaryDownloadUrl {
       name += '1604';
     } else if (majorVer === 16) {
       name += '1604';
+    } else if (majorVer === 18) {
```
AJRdev: Maybe a unit test should be added for this version ?

AJRdev: Maybe we should add a unit test for this case ?

YuriGor: I think we need to merge a fix and publish package ASAP
since it just don't work in Ubuntu 18.04 with default settings.

May be we do not need hard-coded list of versions, LTS Ubuntu versions are always have the same names and mongo binaries files too, so we can just generate right file name for any future releases. 
In this case yes, it will make a sence to add some tests on this function.
But IMHO this will be separate task, more like future proof improvement, not a bug fix.

YuriGor:
May be we need to update default version from 1404 to 1804? 
If somebody uses non-lts distro, it most likely will be the latest one, 
so chances are the latest mongo will fit better.

src/util/MongoBinaryDownloadUrl.js
AJRdev: Well I mean there are already of lot of unit tests of this function in the MongoBinaryDownloadUrl-test.js file, I don't think it will be very long to add the specific case for this version, it would be something like this : 
```javascript
it('should return a archive name for Ubuntu 12.04', () => {
      expect(
        downloadUrl.getUbuntuVersionString({
          dist: 'Ubuntu Linux',
          release: '18',
        })
      ).toBe('ubuntu1804');
```

YuriGor: As @nodkz asked, I [removed checking `release` field](https://github.com/nodkz/mongodb-memory-server/pull/96/commits/eb3c36e1ec408a624cad7fc86bf821a02915182a),
so this test will be failed instantly)

AJRdev: @YuriGor why would it fail ? 
majoVer is constructed by splitting os.release by commas and gathering the first entry : 
```javascript
const ubuntuVer: string[] = os.release ? os.release.split('.') : [];
const majorVer: number = parseInt(ubuntuVer[0], 10);
```
So if you pass this object to the function :
```javascript
{
          dist: 'Ubuntu Linux',
          release: '18',
}
```
majorVer will be equal to 18 so the result will be ubuntu1804 ...

YuriGor: ah, majorVer is parsed from release string, ok, test added.

nodkz:
@YuriGor thanks a lot for PR!

This default value is not so important. It used only when `os` package cannot determine Ubuntu version.

We may change this in another PR. But for now merge changes, cause it helps to solve the current problem for Ubuntu 18 users.

nodkz:
@AJRdev thanks for help!

Issue #95, opened 2018-10-23
Auto-download of binaries on install


guillaumervls:
Hi and thanks for this module,

I use the `downloadDir` options in my code, so my home folder stays clean of `.mongodb-binaries`. However the `postinstall` script is not aware of that and downloads in my home folder. Shouldn't the downloading of binaries after `npm install` be optional ? (I mean opt-in instead of opt-out)

Am I missing something ?

Cheers

AJRdev:
@guillaumervls You should use the environnement variable MONGOMS_DOWNLOAD_DIR that specify a path where to download the mongodb binaries instead of the downloadDir option, because it will be used in the postinstall script as well :)

nodkz:
In some cases, mongod binary may download about 1 minute the first time.

Most developers don't know how to increase timeouts in test runners. And by default after 5 seconds tests halts.

So for making testing without additional setups I decide to download latest verson on postinstall script.It allows to start testing with zero configuration.

nodkz:
Disabling downloading via `MONGOMS_DISABLE_POSTINSTALL` env variable is less evil.

guillaumervls:
Ok then isn't there a more appropriate place to download them (more local to the project) ?

guillaumervls:
Or read a setting from `package.json` ? Setting an environment variable puts this responsibility on the users of the package using this lib, forcing them to use `MONGOMS_DISABLE_POSTINSTALL=1 npm install`. With a setting in `package.json` they could just `npm install`.

nodkz:
Sounds sweet if we can provide some vars via package.json or even .dot file.

Any proposal how it can be implemented? Need to use expierence from jest, babel, ts or any other implementation.

I feel your tension and dissapoinment about current ways of configuration. no one has raised this problem yet and you are the first!  So let's solve this problem.

guillaumervls:
Cool ! From what I've seen popular modules like Husky, Babel, and Jest, use a `package.json` key matching their `npm` name (`babel`, `husky`, `jest`). Only ESLint uses `eslintConfig`. In our case it could be `mongodb-memory-server`.

IMHO, special config files like `.babelrc` should be used if they were so many config option that it could bloat `package.json`. It may not be necessary for the 7 options available here (the 7 `MONGOMS_XXX` env vars).

To implement this, I suggest the use of https://www.npmjs.com/package/pkg-conf which seems quite popular, and does exactly what we want.

nodkz:
[...]

PR #99, opened 2018-11-03
switch to node_modules/.cache directory instead of $HOME for binaries DL

guillaumervls:
Fixes https://github.com/nodkz/mongodb-memory-server/issues/95

### DIFF
--- a/src/util/MongoBinary.js
+++ b/src/util/MongoBinary.js
@@ -1,5 +1,6 @@
 /* @flow */
 
+import fs from 'fs';
 import os from 'os';
 import path from 'path';
 import LockFile from 'lockfile';
@@ -23,10 +24,13 @@ export default class MongoBinary {
   static cache: MongoBinaryCache = {};
 
   static async getPath(opts?: MongoBinaryOpts = {}): Promise<string> {
+    const legacyDLDir = path.resolve(os.homedir(), '.mongodb-binaries');
     const defaultOptions = {
       downloadDir:
         process.env?.MONGOMS_DOWNLOAD_DIR ||
-        path.resolve(findCacheDir({ name: 'mongodb-memory-server' }), 'mongodb-binaries'),
+        (fs.existsSync(legacyDLDir)
+          ? legacyDLDir
+          : path.resolve(findCacheDir({ name: 'mongodb-memory-server' }), 'mongodb-binaries')),
       platform: process.env?.MONGOMS_PLATFORM || os.platform(),
       arch: process.env?.MONGOMS_ARCH || os.arch(),
       version: process.env?.MONGOMS_VERSION || 'latest',
--- a/src/util/MongoBinary.js
+++ b/src/util/MongoBinary.js
@@ -4,6 +4,7 @@ import os from 'os';
 import path from 'path';
 import LockFile from 'lockfile';
 import mkdirp from 'mkdirp';
+import findCacheDir from 'find-cache-dir';
 import MongoBinaryDownload from './MongoBinaryDownload';
 
 export type MongoBinaryCache = {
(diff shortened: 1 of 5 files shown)


AJRdev:
@guillaumervls @nodkz I think there is a problem with the current Travis CI workflow and we have not seen this problem before because until now the cache folder of the mongodb binary was in the **%HOME/.mongodb-binaries** and h'asn't changed apparently in a while.

The Travis CI is failing because it runs the yarn test script before running the yarn build script that transpiles  the files with news changes and runs the post_install script that will download the binary in the new location (node_modules/.cache/mongodb-binaries) and cache it for yarn test script.
Since the binary right now is not cache for the test script, the binary is downloaded during the tests that causes this timeout : 
```bash
Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.
```

Should we have launch the **yarn run build** before the **yarn run test** ?

AJRdev:
@guillaumervls  Oh my bad, I know why it is not working, the new package find-cache-dir wasn't commited into yarn.lock file, so the Travis CI didn't launch the yarn command that would have triggered the post_install script and downloaded the mongodb binaries :)

We just need to do on the branch of the PR, yarn and then commit the changes in yarn.lock to fix the Travis pipeline.

@nodkz And yes, I agree we should fallback on the old %HOME folder until we bump a new major version.

src/util/MongoBinary.js
```diff
@@ -24,7 +25,8 @@ export default class MongoBinary {
   static async getPath(opts?: MongoBinaryOpts = {}): Promise<string> {
     const defaultOptions = {
       downloadDir:
-        process.env?.MONGOMS_DOWNLOAD_DIR || path.resolve(os.homedir(), '.mongodb-binaries'),
```
AJRdev: Like @nodkz suggested we should keep the fallback on **path.resolve(os.homedir(), '.mongodb-binaries')** until a new major version is bumped.

guillaumervls:
@AJRdev @nodkz Done for the fallback. As to yarn.lock, I don't use yarn, could you tell me the update needed ?

src/util/MongoBinary.js
```diff
@@ -23,10 +24,13 @@ export default class MongoBinary {
   static cache: MongoBinaryCache = {};
 
   static async getPath(opts?: MongoBinaryOpts = {}): Promise<string> {
+    const legacyDLDir = path.resolve(os.homedir(), '.mongodb-binaries');
     const defaultOptions = {
       downloadDir:
         process.env?.MONGOMS_DOWNLOAD_DIR ||
-        path.resolve(findCacheDir({ name: 'mongodb-memory-server' }), 'mongodb-binaries'),
+        (fs.existsSync(legacyDLDir)
```
AJRdev: Why did you add this check ?
And not just piped everything like this : 
```javascript
downloadDir:
        process.env?.MONGOMS_DOWNLOAD_DIR ||
        path.resolve(os.homedir(), '.mongodb-binaries') ||
        path.resolve(findCacheDir({ name: 'mongodb-memory-server' }), 'mongodb-binaries')
```

AJRdev:
@guillaumervls For yarn, you just need to install it globally with **npm install -g yarn**, and then run at the root of the project the command **yarn** (the equivalent of npm install), it will install all the dependecies and add the new one in the yarn.lock file, then you just have to commit and push it on your branch.

If you have any warning messages during the yarn command concerning the package-lock.json, just remove then package-lock.json file and re-do the procedure above.

src/util/MongoBinary.js
guillaumervls: `path.resolve(os.homedir(), '.mongodb-binaries')`
is always truthy (it just returns a path), so

`path.resolve(findCacheDir({ name: 'mongodb-memory-server' }), 'mongodb-binaries')`
would never be used

guillaumervls:
Done

guillaumervls:
> @guillaumervls For yarn, you just need to install it globally with **npm install -g yarn**, and then run at the root of the project the command **yarn** (the equivalent of npm install), it will install all the dependecies and add the new one in the yarn.lock file, then you just have to commit and push it on your branch.
> 
> If you have any warning messages during the yarn command concerning the package-lock.json, just remove then package-lock.json file and re-do the procedure above.
> 
> This procedure should fix the Travis CI pipeline.

guillaumervls:
(Sorry, misclicked on close pull request)

nodkz:
@guillaumervls @AJRdev thanks a lot for PR and fantastic work! 🍻👍

All looks good! Merged!

PR #103, opened 2018-11-12
Add check for uppercase HTTP and HTTPS proxy env settings.

deanlot:


AJRdev:
@deanlot Why do you need to add these uppercase env variables since they do exactly the same than process.env.http_proxy and process.env.https_proxy ?

deanlot:
@AJRdev In our environment, some people set http_proxy, and others set HTTP_PROXY. Windows users set HTTP_PROXY, linux users usually set http_proxy.

Also, Our devops team who maintain some of the servers for us, have only set HTTP_PROXY, and to get this changed will take weeks, and lots of effort.

As good practice, I think we should make this support all variations of proxy config. This will save many people a lot of debugging time.

nodkz:
@deanlot thanks! 👍

PR #110, opened 2018-11-27
clean up skipDownload code

benjamingr:
We were trying to figure out why this doesn't work (before 2.7.3 in the postinstall) and found this code confusing so I figured I'd contribute back and clean it up (thanks for the library!)

### DIFF
--- a/postinstall.js
+++ b/postinstall.js
@@ -13,9 +13,8 @@ function isModuleExists(name) {
   catch(e) { return false }
 }
 
-const skipDownload = typeof process.env.MONGOMS_DISABLE_POSTINSTALL === 'string'
-  ? ['1', 'on', 'yes', 'true'].indexOf(process.env.MONGOMS_DISABLE_POSTINSTALL.toLowerCase()) !== -1
-  : false;
+const skipDownload = (typeof process.env.MONGOMS_DISABLE_POSTINSTALL === 'string') &&
+  ['1', 'on', 'yes', 'true'].includes(process.env.MONGOMS_DISABLE_POSTINSTALL.toLowerCase())
 
 if (skipDownload) {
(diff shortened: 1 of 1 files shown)


nodkz:
Sorry man, i cannot acept `includes` method. Cause it will not work for old node versions. We are really need to accept this PR?

For old node versions we should use `indexOf` cause postinstall.js doea not transpilled by babel.

AJRdev:
@benjamingr I don't really understand why you found the code confusing, was it because of the ternary ? And like @nodkz stated we can't use `inclues` in the **postinstall** script since it is not transpilled..

benjamingr:
>  was it because of the ternary ? 

Yes, mostly because it was a confusing way to express && in this example.
 
> And like @nodkz stated we can't use inclues in the postinstall script since it is not transpilled..

What node versions do you support?

> We are really need to accept this PR?

No, of course not - I've found it to be an improvement, this is your project and you call the shots here.

nodkz:
This lib supports node 6 and above.

`includes` does not works for node <6.5
http://kangax.github.io/compat-table/es2016plus/

So I'm afraid to merge this Pull Request for now.
Will be merged with next major version. But when it will happen I don't know.
Let keep this issue opened till a better time.

LoneRifle:
> @benjamingr I don't really understand why you found the code confusing, was it because of the ternary ? And like @nodkz stated we can't use `inclues` in the **postinstall** script since it is not transpilled..

I believe it's just the tenary that can be a bit confusing (ie, why use a tenary to determine the value of a boolean?). If @benjamingr can refine the change so that it uses `indexOf` rather than `includes` that should make this PR mergeable

benjamingr:
Sure

nodkz:
Thanks! Agreed with your changes!

Issue #111, opened 2018-12-02
auth param not passed when instance.auth is true


StefanoDeVuono:
[PR 79](https://github.com/nodkz/mongodb-memory-server/pull/79) doesn't add `auth` to the command line arguments (`prepareCommandArgs()`) even when `instance.auth` is set to `true`.

AJRdev:
@StefanoDeVuono I close this issue, since your PR has been merged, thanks for your contribution !

PR #112, opened 2018-12-02
allow authorizations

StefanoDeVuono:
As referenced in [issue 111](https://github.com/nodkz/mongodb-memory-server/issues/111) PR 79 doesn't add auth to the command line arguments (prepareCommandArgs()) even when instance.auth is set to true.

This PR allows auth to be passed as a command line argument.


nodkz:
Thanks 👍

PR #114, opened 2018-12-06
add MONGOMS_SKIP_MD5_CHECK environment option

dimitriylol:
There are several changes in this pull request:
- add error handling in postinstall.js script for verbose error during downloading and installation
- add MONGOMS_SKIP_MD5_CHECK environment variable for skipping MD5 check with the reference result

FYI: md5 sum has been updated in the MongoDB mirror, so I removed my workaround in MongoBinary-test.js. Even this pull request currently isn't required, I think it's good to have such option for the future.

### DIFF
--- a/src/util/MongoBinaryDownload.js
+++ b/src/util/MongoBinaryDownload.js
@@ -91,6 +91,14 @@ export default class MongoBinaryDownload {
   }
 
   async checkMd5(mongoDBArchiveMd5: string, mongoDBArchive: string) {
+    if (
+      typeof process.env.MONGOMS_SKIP_MD5_CHECK === 'string'
+        ? ['1', 'on', 'yes', 'true'].indexOf(process.env.MONGOMS_SKIP_MD5_CHECK.toLowerCase()) !==
+          -1
+        : false
+    ) {
+      return undefined;
+    }
     const signatureContent = fs.readFileSync(mongoDBArchiveMd5).toString('UTF-8');
     const m = signatureContent.match(/(.*?)\s/);
     const md5Remote = m ? m[1] : null;
@@ -98,6 +106,7 @@ export default class MongoBinaryDownload {
     if (md5Remote !== md5Local) {
       throw new Error('MongoBinaryDownload: md5 check is failed');
     }
+    return undefined;
   }
 
   async download(downloadUrl: string) {
--- a/src/util/MongoBinaryDownload.js
+++ b/src/util/MongoBinaryDownload.js
@@ -18,22 +18,40 @@ export type MongoBinaryDownloadOpts = {
   platform: string,
   arch: string,
   debug?: DebugPropT,
+  skipMD5?: boolean,
 };
 
 export default class MongoBinaryDownload {
   debug: DebugFn;
   dlProgress: DownloadProgressT;
 
+  skipMD5: boolean;
   downloadDir: string;
   arch: string;
   version: string;
   platform: string;
 
-  constructor({ platform, arch, downloadDir, version, debug }: $Shape<MongoBinaryDownloadOpts>) {
+  constructor({
+    platform,
+    arch,
+    downloadDir,
+    version,
+    skipMD5,
+    debug,
+  }: $Shape<MongoBinaryDownloadOpts>) {
     this.platform = platform || os.platform();
     this.arch = arch || os.arch();
     this.version = version || 'latest';
     this.downloadDir = path.resolve(downloadDir || 'mongodb-download');
+    if (skipMD5 === undefined) {
+      this.skipMD5 =
+        typeof process.env.MONGOMS_SKIP_MD5_CHECK === 'string'
+          ? ['1', 'on', 'yes', 'true'].indexOf(process.env.MONGOMS_SKIP_MD5_CHECK.toLowerCase()) !==
+            -1
+          : false;
+    } else {
+      this.skipMD5 = skipMD5;
+    }
     this.dlProgress = {
       current: 0,
       length: 0,
@@ -91,12 +109,7 @@ export default class MongoBinaryDownload {
   }
 
   async checkMd5(mongoDBArchiveMd5: string, mongoDBArchive: string): Promise<?boolean> {
-    if (
-      typeof process.env.MONGOMS_SKIP_MD5_CHECK === 'string'
-        ? ['1', 'on', 'yes', 'true'].indexOf(process.env.MONGOMS_SKIP_MD5_CHECK.toLowerCase()) !==
-          -1
-        : false
-    ) {
+    if (this.skipMD5) {
       return undefined;
     }
     const signatureContent = fs.readFileSync(mongoDBArchiveMd5).toString('UTF-8');
--- a/src/util/MongoBinaryDownload.js
+++ b/src/util/MongoBinaryDownload.js
@@ -90,7 +90,7 @@ export default class MongoBinaryDownload {
     return mongoDBArchive;
   }
 
-  async checkMd5(mongoDBArchiveMd5: string, mongoDBArchive: string) {
+  async checkMd5(mongoDBArchiveMd5: string, mongoDBArchive: string): Promise<?boolean> {
     if (
       typeof process.env.MONGOMS_SKIP_MD5_CHECK === 'string'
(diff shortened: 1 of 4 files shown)


src/util/MongoBinaryDownload.js
```diff
@@ -91,13 +91,22 @@ export default class MongoBinaryDownload {
   }
 
   async checkMd5(mongoDBArchiveMd5: string, mongoDBArchive: string) {
```
AJRdev: I know it's not you that created this function but I don't understand why this function is declared async since there are no asynchronous operation in it.

src/util/MongoBinaryDownload.js
```diff
@@ -91,13 +91,22 @@ export default class MongoBinaryDownload {
...
+    if (
```
AJRdev: This check could be done directly before the call of the function here : 
https://github.com/nodkz/mongodb-memory-server/blob/4f0e7ca32ff0fc90147935d9ff8d6284f755d4d6/src/util/MongoBinaryDownload.js#L88
Like that we will not have to return an undefined.

src/util/MongoBinaryDownload.js
```diff
@@ -91,13 +91,22 @@ export default class MongoBinaryDownload {
...
+      return undefined;
+    }
     const signatureContent = fs.readFileSync(mongoDBArchiveMd5).toString('UTF-8');
     const m = signatureContent.match(/(.*?)\s/);
     const md5Remote = m ? m[1] : null;
     const md5Local = md5File.sync(mongoDBArchive);
     if (md5Remote !== md5Local) {
       throw new Error('MongoBinaryDownload: md5 check is failed');
     }
+    return undefined;
```
AJRdev: Why do we need to return an undefined here ?

src/util/MongoBinaryDownload.js
dimitriylol: @AJRdev me too. Sure, I'll change it

src/util/MongoBinaryDownload.js
dimitriylol: @AJRdev this function normally returns `undefined`, so I make the same explicitly. I don't want to move it out of this function because in the future someone else can write `new MongoBinaryDownload({}).checkMd5('', '')` and MONGOMS_SKIP_MD5_CHECK would be missed.

src/util/MongoBinaryDownload.js
dimitriylol: @AJRdev if you add somewhere in the function body `return ` statement, then eslint doesn't tolerate implicit `return undefined`

src/util/MongoBinaryDownload.js
AJRdev: @dimitriylol Ok cool :) I think it will impact all your tests that expect a Promise from this function.

src/util/MongoBinaryDownload.js
AJRdev: @dimitriylol Well I don't even see why checkMd5 should be a public member function of the MongoBinaryDownload class since it is only used internally in this class. I don't see how it would be useful for other class to call the method checkMd5 of this class, and if it is the case it should not be a member function but more an utility function not linked to a specific class.

dimitriylol: @AJRdev totally agree with you about moving it to a separate module. Waiting for the clarification from @nodkz

src/util/MongoBinaryDownload.js
nodkz: Async `checkMd5` is made for compatibility with other download methods. So better to keep it as is.

Better to rewrite implementation in async way. But let it keep for another PR.

dimitriylol: @nodkz if so, I need to remove my last commit

src/util/MongoBinaryDownload.js
nodkz: Also better to create a new property `skipMD5` for `MongoBinaryDownloadOpts` & class. Then use it inside `checkMd5` method.

And move the reading of env variable to `constructor` if `skipMD5===undefined`:
- it makes tests more simple
- it provides a feature to disable md5 check via config options (don't forget to update README) not only via ENV vars

src/util/MongoBinaryDownload.js
nodkz: We can make improve signature `async checkMd5(mongoDBArchiveMd5: string, mongoDBArchive: string): Promise<?boolean>` and return at the end a `true` value. 

At line 100 keep `undefined`, cause we do not check md5 and not sure "true or false" is it.

src/util/MongoBinaryDownload.js
dimitriylol: @nodkz I'm sorry, I'm not clear about this. Do you want to pass it via constructor parameter *and* via ENV variable? Maybe only one source of truth is preferable. Either constructor parameter or ENV variable.

nodkz: Users may call MMS in different scenarios and may in one process run several servers with different versions. 

So it will be good if we read `skipMD5` from options firstly, and if it does not provided then read from global env variable.

dimitriylol: @nodkz OK, I'll make appropriate changes and add this priority to the documentation

nodkz: From readme:
```diff
const mongod = new MongoMemoryServer({
  instance: {
    port?: ?number, // by default choose any free port
    ip?: string, // by default '127.0.0.1', for binding to all IP addresses set it to `::,0.0.0.0`,
    dbName?: string, // by default generate random dbName
    dbPath?: string, // by default create in temp directory
    storageEngine?: string, // by default `ephemeralForTest`, available engines: [ 'devnull', 'ephemeralForTest', 'mmapv1', 'wiredTiger' ]
    debug?: boolean, // by default false
    replSet?: string, // by default no replica set, replica set name
    auth?: boolean, // by default `mongod` is started with '--noauth', start `mongod` with '--auth'
    args?: string[], // by default no additional arguments, any additional command line arguments for `mongod` `mongod` (ex. ['--notablescan'])
  },
  binary: {
    version?: string, // by default 'latest'
    downloadDir?: string, // by default node_modules/.cache/mongodb-memory-server/mongodb-binaries
    platform?: string, // by default os.platform()
    arch?: string, // by default os.arch()
    debug?: boolean, // by default false
+    skipMD5?: boolean, // by default false OR process.env.MONGOMS_SKIP_MD5_CHECK
  },
  debug?: boolean, // by default false
  autoStart?: boolean, // by default true
});
```

src/util/MongoBinaryDownload.js
dimitriylol: @nodkz sure, we'll do it

src/util/MongoBinaryDownload.js
dimitriylol: @nodkz seems like, I finished. Please take a look

src/util/MongoBinaryDownload.js
```diff
@@ -18,22 +18,40 @@ export type MongoBinaryDownloadOpts = {
...
+    version,
+    skipMD5,
+    debug,
+  }: $Shape<MongoBinaryDownloadOpts>) {
     this.platform = platform || os.platform();
     this.arch = arch || os.arch();
     this.version = version || 'latest';
     this.downloadDir = path.resolve(downloadDir || 'mongodb-download');
+    if (skipMD5 === undefined) {
+      this.skipMD5 =
```
AJRdev: Just to be consistent with a recent merged PR (https://github.com/nodkz/mongodb-memory-server/pull/110/files#diff-551007241ae61889b0f85f47aa764343R16), could you use this syntax instead : 
```javascript
this.skipMD5 =(typeof process.env. MONGOMS_SKIP_MD5_CHECK === 'string') &&
  (['1', 'on', 'yes', 'true'].indexOf(process.env. MONGOMS_SKIP_MD5_CHECK.toLowerCase()) !== -1)
```

AJRdev:
@nodkz Just a little change request for consistency regarding a recent PR but out of that everything seems good to me :)

src/util/MongoBinaryDownload.js
dimitriylol: @AJRdev updated but without parentheses. Eslint removes parentheses in the expression.

nodkz:
Thanks to all for amazing Pull Request! Especially regards to @dimitriylol ;)

Issue #32, opened 2017-11-16
Can't run mongodb-memory-server on Linux Alpine


Kostanos:
I got the error:
```
Cannot determine version string for {"os":"linux","dist":"Alpine Linux","release":"3.4.6"}
```

nodkz:
Can you take a look on https://github.com/nodkz/mongodb-memory-server/blob/master/src/util/MongoBinaryDownloadUrl.js and test for this file.

Will be awsome if you make Pull Request with fixing this error.
Thanks for help!

Kostanos:
I'll check later, temporarily switched to docker image with mongo, need to finish my project first.

bj97301:
@Kostanos What image did you end up going with?

ericbarch:
FYI for anyone running into this issue -- switching from the nodejs alpine image to the nodejs debian stretch image solved this for me (albeit with a MUCH larger image).

I'm getting an ENOENT error with the alpine image. This sums up why: https://stackoverflow.com/questions/32724556/execute-mongodb-binaries-on-alpine-linux

nodkz:
<img width="692" alt="screen shot 2018-08-21 at 21 35 06" src="https://user-images.githubusercontent.com/1946920/44412195-1f0fd000-a58a-11e8-86dc-7cba1a2e9d7e.png">

So if Mongo team makes alpine build we can easily add it to this package
https://www.mongodb.org/dl/linux

An opened issue in MongoDB: https://jira.mongodb.org/browse/SERVER-36790

jloveridge:
At present it seems unlikely that MongoDB is going to release a separate Alpine linux binary themselves. From their [official documentation](https://docs.mongodb.com/manual/administration/install-on-linux/) they direct users to the [Alpine Linux package](https://pkgs.alpinelinux.org/package/edge/community/x86_64/mongodb).

I would suggest that users who want to run their tests inside of an Alpine Linux container. Either add the specifically mentioned packages to their `Dockerfile` used for their tests or base their container on something like the [alpine-mongo](https://hub.docker.com/r/mvertes/alpine-mongo/) container that is already available on [Docker Hub](https://hub.docker.com/).

jtbairdsr:
[...]

PR #119, opened 2018-12-18
System binary

jtbairdsr:


jtbairdsr:
I have pushed a change in response to your comments https://github.com/nodkz/mongodb-memory-server/issues/32#issuecomment-448137640.  I look forward to your feedback...

src/util/MongoBinary.js
```diff
@@ -43,84 +122,60 @@ export default class MongoBinary {
...
-            : `MongoBinary: Download lock removed`
-        );
-      });
+    if (!binaryPath) {
+      binaryPath = await this.getDownloadPath(options);
     }
 
-    debug(`MongoBinary: Mongod binary path: ${this.cache[version]}`);
-    return this.cache[version];
+    if (version !== 'latest' && systemBinary) {
```
nodkz: We should not show the message if versions are the same ;)

And I think that better to move this code inside `if (systemBinary) {` at line 149
right after `binaryPath = await this.getSystemPath(systemBinary);`

nodkz: All rest parts are good! Code became much cleaner! Thanks!

nodkz: Ouch! And don't forget to add new option to README.md

jtbairdsr: Thanks, I appreciate the feedback and agree.  I have updated the code and the README.md.  let me know if there is anything else.

src/util/MongoBinary.js
```diff
@@ -148,6 +148,21 @@ export default class MongoBinary {
 
     if (systemBinary) {
       binaryPath = await this.getSystemPath(systemBinary);
+      const binaryVersion = execSync('mongod --version')
```
nodkz: Should be
```js
const binaryVersion = execSync(`${binaryPath} --version`)
```

And need to add check that `binaryPath` is not empty ;)

jtbairdsr: oh yeah, good catch totally missed that.

nodkz: I've just merged with the master (there was a conflict).
Pull changes before commit.

nodkz: There were updated packages in master.
Maybe it helps to resolve #120

jtbairdsr: 👌 I made that fix and pushed.

nodkz:
Thanks a lot! 👍

nodkz:
I'll fix
const binaryVersion = execSync(`${binaryPath} --version`)
myself

jtbairdsr:
No problem! It was a pleasure to participate.

jtbairdsr:
when will these changes be published to npm?

jtbairdsr:
@nodkz thank you again for the awesome package!  It was a pleasure to contribute!

PR #122, opened 2018-12-22
fix: promisfy fs.access in getSystemPath

dkphung:
Trying to use the MongoBinary on my alpine container and ran into an issue where it can't find the binary file i pass in even though the binary exists.  Trace down the problem and is seem to be related to fs.access() throwing the error where callback is not defined.  fs.access is not a promise so we need to promisify for the code to work.


nodkz:
Thanks a lot!

PR #129, opened 2019-01-10
fix(MongoInstance): kill the `killlerProcess` on instance kill

skrtheboss:
When a `MongoInstance` is killed, the `killerProcess` should also be killed. Else `jest` may fail in other projects, since there remains an `open handle` for the `killerProcess` until the interval (2s)  is called, see  [Mongo Killer Interval](https://github.com/nodkz/mongodb-memory-server/blob/b59ed826dd4f3cd31ad6f05c890f70eaad034abd/src/util/mongo_killer.js#L32).

### DIFF
--- a/src/util/MongoInstance.js
+++ b/src/util/MongoInstance.js
@@ -111,6 +111,12 @@ export default class MongodbInstance {
         this.childProcess.kill();
       });
     }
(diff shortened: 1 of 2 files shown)


nodkz:
Wow good catch! 👌
Tnx!

PR #137, opened 2019-02-14
Fix #136 - Fix MD5 check failing due to validating minimum size for MD5 file itself

### LINKED ISSUE #136: checkMD5 always fails due to recent "too small" validation being done to (small) md5 file
I'm running:

```js
process.env.MONGOMS_MD5_CHECK = 'true';
mongoServer = new MongoMemoryServer({
  binary: {
    version: '3.6.8',
  },
});
```

Download goes just fine, but then mongoms aborts saying ` Error: Too small (76 bytes) mongod binary downloaded from https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1604-3.6.8.tgz`

Except it's not true! Manually checking the tgz downloaded in `~/.mongodb-binaries`, I confirmed the md5. And then,

```
20:14:19 ~/.mongodb-binaries ls -la mongo*
-rw-rw-r-- 1 ronanj ronanj 100520377 Feb 13 20:08 mongodb-linux-x86_64-ubuntu1604-3.6.8.tgz
-rw-rw-r-- 1 ronanj ronanj        76 Feb 13 20:08 mongodb-linux-x86_64-ubuntu1604-3.6.8.tgz.md5.downloading
                                  ^^
                                  It's the *checksum* file that is 76 bytes! :D
```

→ `httpDownload` should skip the new "too small" validation for the md5 file.

Also, I'd like to ask for md5 validation using the `binary: {checkMD5: true}` option (rather than writing to `process.env.MONGOMS_MD5_CHECK`), but it looks like the TS typings don't include this option, so tsc compilation fails.

ronjouch:
Also,

- Fix TS typings to accept documented `binary.checkMD5` flag.

### DIFF
--- a/src/util/MongoBinary.d.ts
+++ b/src/util/MongoBinary.d.ts
@@ -10,6 +10,7 @@ export interface MongoBinaryOpts {
   platform?: string;
   arch?: string;
   debug?: DebugPropT;
+  checkMD5?: boolean;
 }
 
 // disable error for a class with all static functions,
--- a/src/util/MongoBinaryDownload.js
+++ b/src/util/MongoBinaryDownload.js
@@ -223,7 +223,7 @@ export default class MongoBinaryDownload {
         response.pipe(fileStream);
 
         fileStream.on('finish', () => {
-          if (this.dlProgress.current < 1000000) {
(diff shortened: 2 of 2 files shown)


nodkz:
Thanks 👍

PR #143, opened 2019-02-28
check for spaces and wrap in quotes

DrogoNevets:
This fixes an issue we are having using this with windows
also fixes for the user in issue 139

### DIFF
--- a/src/util/MongoBinary.js
+++ b/src/util/MongoBinary.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-/* eslint no-bitwise: ["error", { "allow": ["~"] }] */
+/* eslint no-bitwise: ['error', { 'allow': ['~'] }] */
 
 import fs from 'fs';
 import os from 'os';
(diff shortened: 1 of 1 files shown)


src/util/MongoBinary.js
```diff
@@ -155,6 +155,10 @@ export default class MongoBinary {
     if (systemBinary) {
       binaryPath = await this.getSystemPath(systemBinary);
       if (binaryPath) {
+        if(~binaryPath.indexOf(' ')) {
```
nodkz: Please add `// eslint-disable-line o-bitwise`

And run lint before committing:
`npm run lint`

DrogoNevets:
how do i get this to re run?

DrogoNevets:
@nodkz there you go, sorry about that - ive not used prettier before, just let vsc do it for me. (which caused conflicts)

will look into it though, likes like a good way to enforce coding style

nodkz:
Thanks a lot! 👍

Issue #86, opened 2018-10-07
storageEngine opt of replSet is ignored


sebastian-nowak:
`storageEngine` option on `replSet` seems to be ignored:

```js
new MongoMemoryReplSet({
  replSet: {
    storageEngine: 'wiredTiger'
  }
})
```

It falls back to the default memory storage.

AJRdev:
@nodkz I've tried to found the source of this issue, I might have found it but I don't understand this portion of the code in the function **getInstanceOpts** in the **MongoMemoryReplSet** class : 
```javascript
 getInstanceOpts(baseOpts: MongoMemoryInstancePropBaseT = {}): MongoMemoryInstancePropT {
    const rsOpts: ReplSetOpts = this.opts.replSet;
    const opts: MongoMemoryInstancePropT = {
      auth: !!rsOpts.auth,
      args: rsOpts.args,
      dbName: rsOpts.dbName,
      ip: rsOpts.ip,
      replSet: rsOpts.name,
    };
    if (baseOpts.args) opts.args = rsOpts.args.concat(baseOpts.args);
    if (baseOpts.port) opts.port = baseOpts.port;
    if (baseOpts.dbPath) opts.dbPath = baseOpts.dbPath;
    if (baseOpts.storageEngine) opts.storageEngine = baseOpts.storageEngine;
    this.debug('   instance opts:', opts);
    return opts;
  }
```
Why do we only extract the storageEngine from the baseOpts and not from the rsOpts as well like the args,dbName, etc ?
If I add this line in the construction of the opts : 
```javascript
    const opts: MongoMemoryInstancePropT = {
      auth: !!rsOpts.auth,
      args: rsOpts.args,
      dbName: rsOpts.dbName,
      ip: rsOpts.ip,
      replSet: rsOpts.name,
      storageEngine: rsOpts.storageEngine,
    };
```
The specified storageEngine specified in the constructor of **MongoMemoryReplSet** is then used and not the default one.

AJRdev:
@jloveridge You might have an explanation as well on this behaviour since you have written the majority of the code of the MongoMemoryReplSet class :)

AJRdev:
Issue fixed by the PRs #147 and #145

PR #145, opened 2019-03-01
WIP : Typescript migration

AJRdev:
### Description

So here is my first draft of a full typescript migration regarding the issue https://github.com/nodkz/mongodb-memory-server/issues/142

I still have a couple of tests that are failing to fix but all the migration from Flow is pretty done 😃 

I am still improving some stuff and I will have a couple of questions & suggestions on the actual code that raised while typing everything with pretty strict ts-lint rules.

These issues should be fixed by this PR as well : https://github.com/nodkz/mongodb-memory-server/issues/125 https://github.com/nodkz/mongodb-memory-server/issues/86

### To-do list :
- [X] Replace Flow typing by Typescript typing in all files
- [X] Remove Babel & Flow tooling and replacing it by tsc
- [x] Fix unit tests
- [x] Fix missing typescript types
- [x] Check null & undefined cases
- [x] Suggest some code improvements
- [x] Try migrate to `eslint` from `tslint` (@nodkz)
- [x] Add prettier (@nodkz)
- [x] Improve tsconfig (@nodkz)
- [x] Add flow typing generation at build process (@nodkz)
- [x] Check `debug` with port prefix `Mongo[${data.port}]` in `MongoInstance`
- [x] Add new unit tests (needs @nodkz validation)
- [x] Remove TODOs from PR (@AJRdev )

PR #147, opened 2019-03-08
Include storageEngine option in replica sets

casperstr:
Fixing #86


nodkz:
Thanks!

PS. It's already fixed in TypeScript version see #145

PR #148, opened 2019-03-08
Pass instance ip option to MongoInstance.run

machikoyasuda:
I was running into an issue where I was trying to set the `ip` option but the mongod command that was spawned still had '127.0.0.1' in it. I believe the option is never actually being passed in.


nodkz:
Thanks!

PS. Manually added to a new TypeScript version see #145

Issue #151, opened 2019-03-10
Test teardown : await mongod.stop() promise doesn't resolve in 4.0.1


dovrosenberg:
I'm using the project with jest to do unit testing.  I installed it all yesterday for the first time and ran into an issue where in my global teardown the stop() call promise never resolves.  I downgraded to 3.1.3 and the issue went away.

AJRdev:
@dovrosenberg Can you show your code and try to launch the MongoMemoryServer with the debug option set to true and show us your logs ?
```javascript
const mongod = new MongoMemoryServer({debug: true});
```

dovrosenberg:
The environment, setup, and teardown are here: (https://jestjs.io/docs/en/mongodb) - I literally copied and pasted.  I then added these lines to module.exports in my jest config to point to the files I created:

```
  testEnvironment: './test/unit/utils/mongoEnvironment.js',
  globalSetup: './test/unit/utils/mongoSetup.js',
  globalTeardown: './test/unit/utils/mongoTeardown.js',
```

There's a ton of test code, so I won't put that all here, but the issue occurred even without any tests running.

I'm using Mongoose, so I have:
```
  beforeAll(() => {
    mongoose.connect(
      global.__MONGO_URI__,
      { dbName: global.__MONGO_DB_NAME__ },
    );
  });

  afterAll(() => {
    mongoose.connection.close();
  });
```

The last line logged with 4.0.1:
`Mongo[56154] CLOSE: null`

With 3.1.3 I see a line 
`Mongo[56154] Shutdown MongoDB server on port 56154 with pid 9664`

and 
`Mongo[56154] Removing tmpDir C:\Users\Dov\AppData\Local\Temp\mongo-mem-3140ZRTziDDMhgcO`

that aren't there with version 4.0.1.

Here's the full log from a small run with 4.0.1:
```
> jest test/unit/specs/graphql/resolvers/games/index.test.js --config test/unit/jest.conf.js --reporters default --watch --verbose false --no-cache
Determining test suites to run...Starting MongoDB instance with following options: {"port":56170,"dbName":"b075daec-9372-4026-b70e-98b80aaa803a","uri":"mongodb://127.0.0.1
:56170/b075daec-9372-4026-b70e-98b80aaa803a","storageEngine":"ephemeralForTest","dbPath":"C:\\Users\\Dov\\AppData\\Local\\Temp\\mongo-mem-11100Smcv9Pbj1psx","tmpDir":{"nam
[...]

PR #153, opened 2019-03-12
Transpile mongo_killer.js script

AJRdev:
Resolves issue :  https://github.com/nodkz/mongodb-memory-server/issues/151


AJRdev:
CI is failling because locally when the tests are run (so before transpilation) the mongo_killer.js file is not available, a quick fix to this is to have at the same time in the repo a mongo_killer.js file and a mongo_killer.ts file.

AJRdev:
Doing something like this https://github.com/TypeStrong/ts-node/issues/619 might be better but out of that I don't really see how can pass our local tests without the mongo_killer.js file available before transpilation..

AhmedFat7y:
@AJRdev moving my comment to here

For the tests to pass with minimal intervention, could we mock `childprocess.spawn` to check for filename if it ends with `.ts` to add `ts-node`? Or mock the `MongoInstance._launchKiller` function to do so?

nodkz:
@AJRdev I completely refactor your code ;)
Move `mongo_killer.js` and `postinstall.js` files to `scripts` folder.
And add `scripts` folder to the package.

It much better and simpler solution!

AJRdev:
@nodkz Yes I was just thinking about doing that :p Perfect thanks ! I think we can merge this PR

PR #155, opened 2019-03-13
Support for replSetInitiate settings

fracasula:
Here's a Pull Request to allow additional settings when initializing a replica set:

```js
const replSet = new MongoMemoryReplSet({
    replSet: {
        configSettings: {
            electionTimeoutMillis: 100,
        },
    },
})
await replSet.waitUntilRunning()
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ to be promoted to Primary before the replica set can be successfully used but th
 (see [electionTimeoutMillis](https://docs.mongodb.com/manual/reference/replica-configuration/#rsconf.settings.electionTimeoutMillis)).
 
 It is possible to force the replica set to elect a node as Primary before the default 10 seconds by changing the `electionTimeoutMillis`
-parameters like shown below:
+parameter as shown below:
 
 ```js
 const replSet = new MongoMemoryReplSet({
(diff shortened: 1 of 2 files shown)


sebastian-nowak:
I think this PR should add docs too, otherwise others will not know about this feature

fracasula:
Good point @sebastian-nowak I'll see what I can do :+1:

fracasula:
Here we are https://github.com/nodkz/mongodb-memory-server/blob/7bd4c0d28fce607682312f629286be1a4834c700/README.md#available-options-for-replica-set

nodkz:
@fracasula 👍👍👍

fracasula:
Thanks :slightly_smiling_face:

nodkz:
@fracasula Awaiting another PR, where this option will be by default, but the user may override it if needed.

In such case you may remove example from README, and leave just options info.
And for most people it will work out-of-the-box as the desired behavior.

fracasula:
That sounds like a good idea but I'm not really sure what would be a good default value for the average test environment. I mean 100ms seems to be enough on mine but it's a pretty decent server. It's running straight without virtualization or containerisation on a i7-6700K CPU @ 4.00GHz, SSDs and 32GB of DDR4.

If I try to lower the timeout down to 50ms it doesn't work already. I don't have a normal laptop I can try this on so putting a default there would be purely guessing for me.

Is there any chance you have a slow machine you can run this on? Anything would still be a pretty decent improvement over the default 10000ms anyway but it's probably better to pick up a value that is safe enough (e.g. 500ms)? :sweat_smile:

nodkz:
500ms is the good value

fracasula:
@nodkz https://github.com/nodkz/mongodb-memory-server/pull/156

PR #156, opened 2019-03-13
electionTimeoutMillis default to 500ms

fracasula:


src/MongoMemoryReplSet.ts
```diff
@@ -224,7 +224,9 @@ export default class MongoMemoryReplSet extends EventEmitter {
       const rsConfig = {
         _id: this.opts.replSet.name,
         members,
-        settings: this.opts.replSet.configSettings || {},
+        settings: this.opts.replSet.configSettings || {
+          electionTimeoutMillis: 500,
+        },
```
nodkz: Nope

```js
settings: {
  electionTimeoutMillis: 500,
  ...this.opts.replSet.configSettings
}
```

will be better.

fracasula: Just realised we had all defaults in the constructor so I moved it there. Looks good?

nodkz: I think that my variant will be better.

Assume that all works perfectly for some user with defaults settings. And it decides to set other option for RS. And after that, all his test begin to work on 10 seconds longer.

The user must override this `electionTimeoutMillis` only explicitly, setting up other options should not remove this value.

fracasula: Fixed :+1:

nodkz:
Thanks a lot! 👍
Just saved tons of seconds for testers 👌

Issue #159, opened 2019-03-18
Mongod binary is getting deleted by yarn


BrandonLWhite:
The projects I'm working on use the yarn package manager. We recently began integrating use of mongodb-memory-server with much delight.  However, we've noticed that yarn likes to delete subdirectories and files under node_modules that it determines do not correspond to installed packages.  Because of this, the /node_modules/.cache is being deleted whenever a yarn operation such as `yarn upgrade some-package@latest` is run.

I'm very much trying to avoid mongodb-memory-server needing to download the mongod binary during test execution with the preference being to download it on package installation of mongodb-memory-server.  The initial `yarn install` works fine.  But after that whenever you use yarn to make any changes to packages, the .cache file gets removed and the mongodb-memory-server post install script is NOT run.

In my limited research, it seems that Yarn does not like things writing to node_modules and they are [very opinionated about this stance](https://github.com/yarnpkg/yarn/issues/5709).
I'm wondering if there is a way for mongodb-memory-server to indicate that the mongod binary belongs to the package so as to prevent yarn from pruning it.

nodkz:
According to this issue we should drop using `find-cache-dir` package and migrate on some another solution.

How do you think better to store cache somewhere globally or inside node_modules?

Related issue
https://github.com/nodkz/mongodb-memory-server/issues/95

AJRdev:
@BrandonLWhite as a quick fix, couldn't you use the environment variable **MONGOMS_DOWNLOAD_DIR** to specify a binary download directory instead of the default /node_modules/.cache directory ?

BrandonLWhite:
@AJRdev We avoid requiring special environment setups.  `yarn && yarn test` for example, must do everything needed for developers and CD/CI containers (eg Jenkins).  If you know of a way to set an environment var from a package please let me know!
I am using mongodb-memory-server in a private package library which is then a devDependency of dozens of other packages.  My goal is to manage all mongodb-memory-server settings from within my private package.
Out of scope for this ticket/issue, but I would like to see a way to configure the mongodb-memory-server binary from within the package.json of another package.  I hope to be able to look into this soon.  

The dream would be:

- Package myTestLib has mongodb-memory-server as a dependency.
- Package myTestLib contains a parameter for the mongod binary version, and possibly the download destination directory, in its package.json.
[...]

PR #167, opened 2019-04-10
feat(install): configure installation from package.json

kjarmicki:
Hi :)

I'd like to be able to install desired version of Mongo binary without having to prefix `npm install` with `MONGOMS_VERSION` and other env variables, and without slowing down my first run of tests. I think ability to declare binary installation options in package.json could be useful in this case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,8 @@ const mongod = new MongoMemoryServer({
 });
 ```
 
-Also you can use the environment variables for configure installation process
+You can also use the environment variables or package.json's `config` section to configure installation process.
+Environment variables have higher priority than contents of package.json.
 
 ```txt
 MONGOMS_DOWNLOAD_DIR=/path/to/mongodb/binaries
@@ -96,6 +97,22 @@ MONGOMS_DISABLE_POSTINSTALL=1 # if you want to skip download binaries on `npm i`
 MONGOMS_SYSTEM_BINARY=/usr/local/bin/mongod # if you want to use an existing binary already on your system.
 MONGOMS_MD5_CHECK=1 # if you want to make MD5 check of downloaded binary.
 # Passed constructor parameter `binary.checkMD5` has higher priority.
+```
+
+```json
+"config": {
+  "mongoDbMemoryServer": {
+    "downloadDir": "/path/to/mongodb/binaries",
+    "platform": "linux",
+    "arch": "x64",
+    "version": "3",
+    "debug": "1",
+    "downloadMirror": "url",
+    "disablePostinstall": "1",
+    "systemBinary": "/usr/local/bin/mongod",
+    "md5Check": "1"
+  }
+}
 ```
 
 ### Replica Set start:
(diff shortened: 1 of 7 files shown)


kjarmicki:
Wow, some things have seriously changed in this project structure since yesterday :open_mouth: 
Nevertheless, let me know if that PR makes sense to you and I'll try to adjust it to new project structure.

nodkz:
Configure from package.json is the desired feature  (https://github.com/nodkz/mongodb-memory-server/issues/159#issuecomment-474706911). So, of course, your PR make sense.  👍

For now I see just one downside, that your config option is called `mongoDbMemoryServer`, existed ENV variables called `MONGOMS_*` and current lib has name `mongodb-memory-server` (in memory it by default, but can be used other storage engines).

So before introducing your feature may be better to rename this package. And give it to something universal?!

Do you have any ideas about how the package(s) can be named?

Restrictions:
- short name
- one-word
- same naming for packages, envs vars and config in package.json

kjarmicki:
I think the current name is great actually :slightly_smiling_face: it made me find this package quick and it does exactly what the name says. I don't think one-word name would carry enough meaning to achieve the same effect.
You could try something like `mongo-launcher`, but again, I think current name is not the problem.

I agree that naming consistency is importnant, so maybe `MONGOMS_*` vars could be replaced by `MONGODB_MEMORY_SERVER_*`? Other names seem to be consistent.

On the side note, why did you introduce Lerna? It seems that subpackages are based on the Mongo version, which is just a configuration option, and most of the code is duplicated?

nodkz:
Lerna allows publishing several packages from one repo. Several packages need to provide specific behavior and split lib into more granular parts (e.g., move replica set to another package, cause it requires `mongodb` package for reading RS state and now in `optionalDependencies`. And download `mongodb` driver package for 95% of users who do not use it is quite strange.). 

In other words, splitting into packages will allow building more great tooling around MongoDB.

kjarmicki:
Agree about replicaSet :+1: I just thought about `mongodb-memory-server-global-3.4`, `...-3.6` and `...-4.0` as being a bit redundant, since they only seem to use one different configuration option and share most of the code.

nodkz:
Yep, code reused. Differs only in config options. But such packages remarkably simplify library setup with zero configs.

And when you list packages, you understand what happens without digging configs. It just helpers not more.

Anyway, the core package must have the ability to be configured via package.json.configs.XXX.options.

Needs day or two for choosing a new package name.

kjarmicki:
@nodkz should I backport my PR to the new project structure or do you feel like I should wait a bit more because there are other structural changes coming?

nodkz:
Backport! And it will be shipped.

Upcoming changes only hit config key name. And it will not be a problem, cause all names will be charged to new one.

nodkz:
Backport! And it will be shipped.

Upcoming changes only hit config key name. And it will not be a problem, cause all names will be charged to new one.

kjarmicki:
Hey, sorry for the delay, sometimes life happens. I haven't forget about this PR :)

PR #169, opened 2019-04-14
fix(downloadUrl): elementaryOS support fixed for versions up to 0.3

goooseman:
It used to download the release for ubuntu 14.04 for any elementaryOS.
But different elementaryOS versions use different ubuntu base version.

So there is a need to check ubuntu versuin first and then use it to generate archive name.


nodkz:
Thanks a lot 👍

PR #170, opened 2019-04-14
Elementary os fixed

goooseman:
An addition to #169
Forgot to put extra `.trim()` to take out the `\n` in the end of stdout from the command. Now is fixed.


nodkz:
Thanks a lot 👍

PR #176, opened 2019-04-21
feat(install): configure installation from package.json

kjarmicki:
Continuation of #167 
I've tested it best to my ability, but it was a bit more complicated than previously due to multiple modules. Please check before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ MONGOMS_MD5_CHECK=1 # if you want to make MD5 check of downloaded binary.
 
 ```json
 "config": {
-  "mongoDbMemoryServer": {
+  "mongodbMemoryServer": {
     "downloadDir": "/path/to/mongodb/binaries",
     "platform": "linux",
     "arch": "x64",
--- a/README.md
+++ b/README.md
@@ -110,7 +110,8 @@ const mongod = new MongoMemoryServer({
 });
 ```
 
-Also you can use the environment variables for configure installation process
+You can also use the environment variables or package.json's `config` section to configure installation process.
+Environment variables have higher priority than contents of package.json.
 
 ```txt
 MONGOMS_DOWNLOAD_DIR=/path/to/mongodb/binaries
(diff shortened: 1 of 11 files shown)


packages/mongodb-memory-server-global-3.6/postinstall.js
```diff
@@ -1,22 +1,28 @@
...
   process.env.MONGOMS_DOWNLOAD_DIR = require('path').resolve(
     require('os').homedir(),
     '.cache',
     'mongodb-binaries'
   );
 }
-if (!process.env.MONGOMS_VERSION) {
+if (!resolveConfig('VERSION')) {
   process.env.MONGOMS_VERSION = '3.6.12'; // don't use `latest` it's nightly build
 }
```
nodkz: I think that better will be to create some method in `resolve-config` which setup default variables.

So 
```js
const resolveConfig = require('../mongodb-memory-server-core/lib/util/resolve-config').default;

if (!resolveConfig('VERSION')) {
  process.env.MONGOMS_VERSION = '3.6.12'; // don't use `latest` it's nightly build	
}	}
```

can be rewritten on something more clear:
```js
const setDefaultValue = require('../mongodb-memory-server-core/lib/util/resolve-config').setDefaultValue;

setDefaultValue('VERSION', '3.6.12'); // don't use `latest` it's nightly build	
```

`setDefaultValue` just for example, maybe you can suggest a better name.

README.md
```diff
@@ -125,6 +126,22 @@ MONGOMS_MD5_CHECK=1 # if you want to make MD5 check of downloaded binary.
 # Passed constructor parameter `binary.checkMD5` has higher priority.
 ```
 
+```json
+"config": {
+  "mongoDbMemoryServer": {
```
nodkz: According to the naming of package `mongodb-memory-server` better will be to use the following key `mongodbMemoryServer` (lowercased D).

nodkz:
Amazing! Thanks a lot 👍

nodkz:
Published 5.1.0 on NPM 🚀

kjarmicki:
Woohoo :) Thank you! :tada:

BrandonLWhite:
Thanks @kjarmicki! I'm looking forward to trying this out!

BrandonLWhite:
Can someone provide guidance on how I can configure the downloadDir to be %HOME/.cache from inside my package.json in a platform-agnostic way?

I basically want to achieve this:
`require('path').resolve(require('os').homedir(), '.cache', 'mongodb-binaries') ` but from package.json. 

Is this possible?

nodkz:
@BrandonLWhite use another package https://www.npmjs.com/package/mongodb-memory-server-global

PR #185, opened 2019-05-10
Use cross-spawn to help with Windows issues.

### LINKED ISSUE #180: More Windows space in path issues
This is a follow up from #139.

I thought the fix for that issue solved it, but when I tested again today. I got a new error:

```
Mongo[53261] MongodbInstance: is failed: Error: spawn "C:\Path to Binaries\bin\mongod.EXE" ENOENT +14ms
(node:14412) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'toString' of undefined
    at MongodbInstance._launchKiller (C:\Path to Project\my-project\node_modules\mongodb-memory-server\lib\util\MongoInstance.js:152:148)
    at C:\Path to Project\my-project\node_modules\mongodb-memory-server\lib\util\MongoInstance.js:105:35
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (C:\Path to Project\my-project\node_modules\mongodb-memory-server\lib\util\MongoInstance.js:16:103)
    at _next (C:\Path to Project\my-project\node_modules\mongodb-memory-server\lib\util\MongoInstance.js:18:194)
```

I think the issue might be related to [this node.js bug](https://github.com/nodejs/node/issues/7367#issuecomment-229721296) which was not fixed. The workaround they mention is to quote the path to the command (already done in #139) and add `shell: true` to the spawn options.

I tried adding it [here](https://github.com/nodkz/mongodb-memory-server/blob/v5.1.0/packages/mongodb-memory-server-core/src/util/MongoInstance.ts#L157) and it solved the problem, but I don't know if that's a good idea to do in all cases.

Maybe something like [cross-spawn](https://www.npmjs.com/package/cross-spawn) would be a better option?

dominic-p:
Fixes #180

This is a work in progress because I honestly couldn't figure out how to test it locally with the combination of tools that I don't really use: lerna + typescript + yarn.

Is there a contributing guide somewhere that I missed, or maybe you walk me through how to test locally?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [5.1.2](https://github.com/nodkz/mongodb-memory-server/compare/v5.1.1...v5.1.2) (2019-05-16)
+
+**Note:** Version bump only for package lerna-monorepo
+
+
+
+
+
 ## [5.1.1](https://github.com/nodkz/mongodb-memory-server/compare/v5.1.0...v5.1.1) (2019-05-09)
 
 
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "5.1.1",
+  "version": "5.1.2",
   "command": {
     "publish": {
       "allowBranch": [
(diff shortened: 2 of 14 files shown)


AJRdev:
@dominic-p How to test locally is not really specific to this project, it is the same for every npm package development :-) 

Usually you use a test application that you link to your local package ( this link might help : https://dev.to/vcarl/testing-npm-packages-before-publishing-h7o ) to test your new features.

You can also use in the repo the command **yarn test** to check if your changes didn't break any tests.

dominic-p:
Thanks for the tip. I'm familiar with testing basic npm packages, but I've never worked with lerna or yarn workspaces before. I'll give `npm pack` a try and let you know when I have some time.

PR #194, opened 2019-06-10
Simplify postinstall path check

### LINKED ISSUE #168: Installation process saves binary in wrong location
In version 5.0.1, when the package is installed the binary is not downloaded to the default location (_./node_modules/.cache/mongodb-memory-server/mongodb-binaries_) but to _./node_modules/mongodb-memory-server/node_modules/.cache/mongodb-memory-server/mongodb-binaries_ instead.
When used for the first time then, it will download the binary to the correct location (and the test will probably time out).

Oddly enough, if skipping installation with `MONGOMS_DISABLE_POSTINSTALL=1` and executing `node ./node_modules/mongodb-memory-server/postinstall.js` afterwards (after enabling the installation, that is), the binary is downloaded to the right location.

Here is the installation output when run in Windows:

```bash
{project_root}$ npm install mongodb-memory-server

> mongodb-memory-server@5.0.1 postinstall {project_root}\node_modules\mongodb-memory-server
{project}> node ./postinstall.js

mongodb-memory-server: checking MongoDB binaries cache...
mongodb-memory-server: binary path is {project_root}\node_modules\mongodb-memory-server\node_modules\.cache\mongodb-memory-server\mongodb-binaries\4.0.5\mongod.exe
+ mongodb-memory-server@5.0.1
added 4 packages from 3 contributors in 51.462s
```

Ideally, setting the version in either environment (or _package.json_ in the future) will download the binary in the right location at install time.

dahlbyk:
Fixes #168. The RegExp check is failing with Windows path separators because `\` is treated as a RegExp escape sequence.

I'm not able to get the project to build/run, but I'm pretty sure the logic is equivalent.

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/MongoBinary.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoBinary.ts
@@ -108,7 +108,7 @@ export default class MongoBinary {
 
     // if we're in postinstall script, npm will set the cwd too deep
     let nodeModulesDLDir = process.cwd();
-    while (new RegExp(`node_modules${path.sep}mongodb-memory-server`).test(nodeModulesDLDir)) {
+    while (nodeModulesDLDir.endsWith(`node_modules${path.sep}mongodb-memory-server`)) {
       nodeModulesDLDir = path.resolve(nodeModulesDLDir, '..', '..');
     }
 


nodkz:
Thanks a lot! 👍

PR #191, opened 2019-06-05
fix: resolve to config from closest package.json

czr:
Fix for #187.

This also introduces reInitialisePackageJson() for test scripts. (There might be a cleaner way of doing this with `jest.resetModules()`, but I couldn't figure out how to make it work with TypeScript/ES6 imports.)

It's worth pointing out that if you have a monorepo such as the structure below and you run jest from the project directory then `subproject/test.js` will pick up the config from `project/package.json` whereas if you run jest from inside the subproject directory (or use a tool like lerna which does the same thing) then it will pick up config from `subproject/package.json`. This seems fairly logical to me, and probably as good as we can get, but perhaps others might disagree.

```
project
|-- subproject
|   |-- package.json
|   |-- test.js
|   +-- ...
|-- package.json
+-- ...
```

Note also that this change will break things for anyone who runs tests inside the subproject but has defined config in `project/package.json` to work with the original implementation. (Another possibility would be to walk down the tree of projects and stop at the first one that defines `config.mongodbMemoryServer`.)

If you think these issues need more work, I'm happy to continue working on this PR.

PR #197, opened 2019-06-17
fix: re-initialise configuration using INIT_CWD in postinstall.js

czr:
Fix for issue spotted by @kjarmicki with https://github.com/nodkz/mongodb-memory-server/pull/191

This makes reInitializePackageJson() in resolve-config.ts into a semi-public interface, used by postinstall.js, taking an optional directory argument to search from.

I'm not 100% happy with this solution. I feel like it's a poor interface to have singleton config that can be poked at in this way. It opens up the possibility of action-at-a-distance. Perhaps resolve-config itself should know about INIT_CWD. That would at least make its behaviour more predictable, at the cost of making it somewhat dependent on the details of npm. As always, I'm open to better suggestions.


nodkz:
Seems legit. 👍
For now, I cannot suggest a better solution.

Thanks for PR 🙏

PR #203, opened 2019-07-23
Addressing #74 - Wait until primary is transitioned from secondary for replicaSet

jardakotesovec:
My attempt to has deterministic implementation for waitUntilRunning for replicaSet, which is discussed in #74 .

Was testing if there is way to tell if its ready from `replSetGetStatus`, but my testing shows it rather shows the intention (whats going to be master) rather than actual real time status. With debug info from mongod - you can see that mongod instance is transitioning from secondary state to primary state (after voting) which takes about 2s. Also my testing shows that you get connect error until this transition is finished. And only reliable approach I found was using `serverStatus` which indicates in realtime whether its master or secondary.

Obviously `serverStatus` gives me status for the instance I am connect to, which is first server. Therefore this strategy would fail if some other instance would decide to be master initially. But in practice (testing 3 instances) always first server ended being primary, therefore this approach seems to be reliable from my testing.

Another small change was just rely on `MongoMemoryServer.start` to be resolved (which should resolve once instance starts listening) instead of waiting 1s. That shaved off about 800ms. 

On my computer (current macbook pro) I am getting pretty good times.  waitUntilRunning() gets resolved for single instance: 2100ms and multiple instance (3): 4100ms.

Let me know if you think I missed something. Thanks!

### DIFF
--- a/packages/mongodb-memory-server-core/src/MongoMemoryReplSet.ts
+++ b/packages/mongodb-memory-server-core/src/MongoMemoryReplSet.ts
@@ -173,9 +173,9 @@ export default class MongoMemoryReplSet extends EventEmitter {
       const server = this._startServer(this.getInstanceOpts({}));
       servers.push(server);
     }
+    // ensures all servers are listening for connection
+    await Promise.all(servers.map((s) => s.start()));
     this.servers = servers;
-    // Brief delay to wait for servers to start up.
-    await new Promise((resolve) => setTimeout(resolve, 1000));
     await this._initReplSet();
   }
 
@@ -251,7 +251,7 @@ export default class MongoMemoryReplSet extends EventEmitter {
 
   _startServer(instanceOpts: MongoMemoryInstancePropT): MongoMemoryServer {
     const serverOpts: MongoMemoryServerOptsT = {
-      autoStart: true,
+      autoStart: false,
       debug: this.opts.debug,
       binary: this.opts.binary,
       instance: instanceOpts,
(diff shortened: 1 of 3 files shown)


jardakotesovec:
Alternative approach would be to actually try to connect to replicaset until it does not return error. But this seems to work well.. have ran tests many times..

nodkz:
@jardakotesovec brilliant 👍

You unlocked the ability for easy testing of transactions for the rest of the world 😉

nodkz:
Published on npm as v5.1.8 🎉

nodkz:
Just got some problems on circleCI test.

As fast fix I added a check for infinite loop in `_waitForPrimary` method with following error message  `No PRIMARY elected yet. Timeout expired. Exiting...`.

See last failed CircleCI builds: 
https://circleci.com/gh/nodkz/mongodb-memory-server/tree/master

1) Last test with infinite loop hotfix: https://circleci.com/gh/nodkz/mongodb-memory-server/143
2) Infinite loop problem before hotfix https://circleci.com/gh/nodkz/mongodb-memory-server/138
3) Failed test in repSet test suite https://circleci.com/gh/nodkz/mongodb-memory-server/136

@jardakotesovec will be cool if you have enough time to investigate this problem. Thanks.

jardakotesovec:
@nodkz I can check it out tomorrow - but it seems that you have circleci running only on master? That will make difficult to investigate properly. I can only guess that the first server did not win voting..  so I could probably fallback to strategy really just trying to connect to replicaset until it succeed.

jardakotesovec:
@nodkz actually I could detect that primary is ready&present from stdout message 'transition to primary complete; database writes are now permitted'
Will try that tomorrow. For starting and detecting that mongod is ready is used very much the same strategy and that works well :-).

PR #206, opened 2019-07-24
Possible more robust wait for primary for replica set

jardakotesovec:
Here is quick attempt to improve #203 

Basically detecting from stdout that one of the server transitioned to primary, which seems to be good enough indicator to connect to it.

Sorry for bit misusing typescript. Wanted to quickly try this attempt and typescript takes additional time&effort to make it happy.


jardakotesovec:
Better verison on its way, will open another PR

PR #212, opened 2019-08-08
refactor: 💡 download binary from the full address of mirror url

RodrigoTopan:
MONGOMS DOWNLOAD_MIRROR environment variable should be able to install
binaries from wherever users wants to .Without this refactor, this
variable is used just to allow user to install binaries from another
domain, but this is not considering the full path of binary


nodkz:
@RodrigoTopan, for now, I cannot accept your pull request, because it may break somebody apps.

Please leave logic `DOWNLOAD_MIRROR` as is, and add a new ENV variable `DOWNLOAD_URL`. And add proper changes to README.md:

```diff
- MONGOMS_DOWNLOAD_MIRROR=url # your mirror url to download the mongodb binary
+ MONGOMS_DOWNLOAD_MIRROR=host # your mirror host to download the mongodb binary
+ MONGOMS_DOWNLOAD_URL=url # full URL to download the mongodb binary
```

Thanks!

RodrigoTopan:
Of course. That makes sense. I will make these changes

Thanks!

RodrigoTopan:
New env variable DOWNLOAD_MIRROR added and description made in README.MD file

nodkz:
Thanks a lot! 👍

PR #214, opened 2019-08-28
Migrate to vision media debug

davidecavaliere:
Hello there,

I've create this PR in other to use [visionmedia's debug](https://github.com/visionmedia/debug) instead of console.log. 

With this switch it will be enough to set `DEBUG=mongodb-memory-server:*` when you call your scripts, i.e.:
```
DEBUG=mongodb-memory-server:* yarn watch
```

Any feedback is very welcome. 

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,6 @@ const mongod = new MongoMemoryServer({
     dbName?: string, // by default generate random dbName
     dbPath?: string, // by default create in temp directory
     storageEngine?: string, // by default `ephemeralForTest`, available engines: [ 'devnull', 'ephemeralForTest', 'mmapv1', 'wiredTiger' ]
-    debug?: boolean, // by default false
     replSet?: string, // by default no replica set, replica set name
     auth?: boolean, // by default `mongod` is started with '--noauth', start `mongod` with '--auth'
     args?: string[], // by default no additional arguments, any additional command line arguments for `mongod` `mongod` (ex. ['--notablescan'])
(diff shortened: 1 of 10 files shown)


davidecavaliere:
hello there. any feedback on this please?

nodkz:
I'm very sorry for the delay. 🙏
I wanted to check before my vacation your proposed changes but due bandwidth I didn't can to do this. And after vacation completely miss your PR from my inbox. Shame me. 😉

1) Need to use [debug](https://www.npmjs.com/package/debug) and [@types/debug](https://www.npmjs.com/package/@types/debug) packages instead of `@microgamma/loggator`. Sorry I cannot accept your package due security reasons. Can we use `debug` packages directly?

2) To avoid breaking changes (cause somebody may use `debug` config options) can we return back these options and use [debug.enable()](https://github.com/visionmedia/debug#set-dynamically) method when the set to true?  In such case, we start providing two ways for enabling debug messages - via config options and ENV variable.

3) Need to add info about enabling debugging via ENV var to README.md and return back "old" config options.

Sorry again for delay.

davidecavaliere:
Hello Pavel,

thanks for your reply. No worries, we're all busy with our real work and
our real life, I know ;)

Agree with your points. Will do.

Best,
Davide

hasezoey:
@davidecavaliere are you still planning to continue this PR?

one suggestions: maybe change to `loglevel`? (like in typegoose)

davidecavaliere:
Hello,

sorry for the late reply. Unfortunately I will not be able to update my PR.

hasezoey:
@nodkz because this PR is outdated and @davidecavaliere cant work on this anymore, should we close this an try in another PR with `loglevel`?

nodkz:
@hasezoey I don't feel that this feature is very necessary. MMS does not run in production where pluggable or ENV-configurable log/debug is very necessary.

So let close this PR. 

BUT if somebody wants to realize configurable logs - new PR is welcome ;)

PR #222, opened 2019-09-11
clearTimeout when waiting for primary in replete

iamdoron:
Hi!

My tests were hanging even after stopping the replica set, so I investigated why - turns out there is an open timer that is not cleared


nodkz:
Enabling CircleCI tests for pull requests

nodkz:
Thanks 👍

PR #223, opened 2019-09-13
ReplSetOpts.dbName is optional

mathieug:
ReplSetOpts.dbName is optional (https://github.com/nodkz/mongodb-memory-server/blob/master/packages/mongodb-memory-server-core/src/MongoMemoryReplSet.ts#L70)


nodkz:
Thanks! 👍

PR #224, opened 2019-09-28
Using "useUnifiedTopology" to avoid deprecation msg

Negan1911:
Currently, when running my tests I receive a lot of deprecation notices (see attached img), looks like MongoDB is deprecating his current topology system so they need the `useUnifiedTopology` flag set, i'm sending this PR to correct that, I also upgraded `@types/mongodb` since it was needed for this support.

image

PS: this project is great, keep it up y'all :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "@types/lockfile": "^1.0.1",
     "@types/md5-file": "^4.0.0",
     "@types/mkdirp": "^0.5.2",
-    "@types/mongodb": "^3.1.28",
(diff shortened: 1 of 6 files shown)


nodkz:
Thanks a lot!

Can you please also update package versions in package.json? I try to keep in sync min required versions in package and lock files. It helps to avoid problems when you need to regenerate yarn.lock.

Negan1911:
I just commited the update of the package.json and yarn.lock resolution :)

PR #230, opened 2019-10-04
Clarify readme wording and auto-download behavior

cx1111:
Thanks for this package.

I wanted to clarify some auto-download behavior which wasn't clear in the README. For instance, I assumed that the core package never tries to auto-download the binary. But that is only true upon installation.

Hopefully this makes things clearer for other users.


nodkz:
Thanks a lot! 👍

PR #231, opened 2019-10-08
Improve ReplicaSet example with mongoose code

arodriges:
I think it will be very useful for further googlers


nodkz:
Thank you very much 👍

PR #246, opened 2019-12-03
Compatibility with debian 10

sventschui:
Debian 10 uses libcurl v4, thus needs a different mongodb build


nodkz:
Thanks!

nodkz:
According to https://www.mongodb.org/dl/linux

For Debian 10 it will work only for 4.2.1 and above. For lower mongodb versions there are no builds.

PR #250, opened 2019-12-10
Add support for Linux Mint

hasezoey:
this PR adds support for Linux Mint >= 17 (ubuntu > 1404)
and makes use of `promisify` instead of an wrapper

Test Note: sorry, i couldnt get the tests to run fully AND the eslint error are not made by this pr

Note: When wanting to avoid git dependencies, i could publish it to npm

---

note to the `getos` change:
because the version 3.1.2 didnt get released since August 2 2019 AND because npm dosnt play well with symlinks (https://github.com/retrohacker/getos/pull/98)

---

@nodkz this should fix the libcurl4 issue i mentioned on discord a while ago (until now ubuntu1404 was always used, which was build against libcurl3) - at least i hope it does


nodkz:
Thanks a lot!

For security reasons I cannot accept `"getos": "github:hasezoey/getos#symlink"` (code may be changed in any time). Frankly, for me such dependency is ok, but for somebody, such dependencies may be prohibited.

Lets await day or two, maybe your PR https://github.com/retrohacker/getos/pull/98 for `getos` will be merged and published. If not, then you may publish your own package.

hasezoey:
> Lets await day or two, maybe your PR retrohacker/getos#98 for getos will be merged and published. If not, then you may publish your own package.

ok, thanks for the response, i just thought because 4 other PR's were there and version 3.1.2 didnt got released since August 3, that it would be the only way (this git or a re-published version on npm) for a long while

nodkz:
@hasezoey nothing happens with getos 😞

A found another good solution:
Can you create a folder `getos` with files of your forked repo in https://github.com/nodkz/mongodb-memory-server/tree/master/packages/mongodb-memory-server-core/src/util ?

In such case we avoid git dependency and fix getos package inside this package.

hasezoey:
@nodkz yes could do - but are you sure you would want that?

hasezoey:
@nodkz ok, so basicly i have done it (would just need to apply lint fixed), but before i would do that, i would want to ask if i should change anything - like converting it directly to typescript?

hasezoey:
in addition to the comment above, i mean with "change anything" some of these things:
- completly rewrite to typescript (to remove the .d.ts file) (im currently doing that)
- should re-do how the call is done (no callback anymore, only async/await if needed)
- no dynamic import of these "logic" files anymore (would play better with typescript, would be easier to debug, would just add to type-safety)

nodkz:
As you wish! You may:
A) migrate on typescript
B) leave code as is and add getos to .eslintignore

Both variants are acceptable for solving `mint` problem 😉

hasezoey:
> Both variants are acceptable for solving mint problem wink

i would at least fix the lint problems, migrate to .ts and change the "dynamic" parsing of files to "static"

hasezoey:
@nodkz i thought about it, and got to a solution: using only 4 functions, i could shrink getos by all files to one - the only requirement for the systems are: either support `lsb_release -a` or have an `/etc/os-release` file

could you check out c01bf39224a16035ac4585cb031808d8673a3c79 and test if anything would be wrong? - it is currently in a state for testing the new, i still have all the old code commented out and all files that are not needed for the new are not included (for now)

hasezoey:
i added try-catch around "lsb_release" and cleaned up

ran some tests (with docker):

| Docker Image            | lsb_release | os-release | *-release | Minimal Stage |
|-------------------------|-------------|------------|-----------|---------------|
| alpine:latest           | false       | true       | true      | 2             |
| ubuntu:latest           | false       | true       | true      | 2             |
| opensuse/leap           | false       | true       | true      | 2             |
| opensuse/tumbleweed     | false       | true       | true      | 2             |
| centos:latest           | false       | true       | true      | 2             |
| archlinux:latest        | false       | false      | true      | 3             |
| linuxmintd/mint19-amd64 | true        | true       | true      | 1             |

(take the `getLinuxInfomation` comments to help to know what stages mean)

hasezoey:
@nodkz could you help me getting yarn again, accidentally used npm to install, tried with `yarn install` to get to the new state, but it fails with "Cannot find module './getos'"

nodkz:
It's a very interesting error. I met this problem the first time. 
If you look inside package.json you will see such section
```
"workspaces": {
    "packages": [
      "packages/*"
    ]
  },
```
It informs yarn that repo is a monorepo. And when installing packages then link all "packages/*" folders to node_modules. So after installin packages, after linking monorepo folders; it run postscripts from all packages. And I think, that you have old built scripts in `packages/mongodb-memory-server-core/lib` which use `getos`. But `getos` already removed.

So the problem can be solved in two ways:
1)  `cd packages/mongodb-memory-server-core && yarn build && yarn install`
2) OR `yarn install --ignore-scripts`

------

I resolved this problem and fix all eslint errors.
We are ready to merge with master and publishing a new version?

hasezoey:
i just removed all node_modules folders for any sub-folder in the repo and ran `yarn install` at the root, still the same error

PS: i never worked with a mono-repo before

hasezoey:
Update: i needed to delete `lib` in `mongodb-memory-server-core` too - why didnt this get overwritten?

nodkz:
Really I don't know why it not overwritten.

Just pull my changes and run `yarn install --ignore-scripts && yarn build`

hasezoey:
> Really I don't know why it not overwritten.

like i said, i just had to clean `lib` for `-core`, and it got re-generated, but not overwritten when it already existed

---

i would say that this PR is now ready to merge, even though most of the test still dont run on my machine (i still dont know why)

PS: i found that in `MongoInstance-test` `/data` is used instead of `tmpDir.name` but i dont know if this is intended (and some `debug: true` options are set there too)

PPS: did some testing, every test below (included the said) `'should start instance on port 27333'` does just keep the tests running infinitely (never finishes) [probably because all test below & the listed are "running" mongodb, the tests before are just calling the instance, without running it]

nodkz:
Can you share your tests' output or names of tests which running infinitely?
And what OS name+version do you use? I try to debug them via docker container tomorrow.

hasezoey:
ok, could figure it out:
- first i got an error `CURL_OPENSSL_3` not found, didnt know what to do
- tried deleting `~/.cache/mongodb-binaries`, which fixed it by downloading the correct mongo version for linux

TL;DR: i implemented Linux Mint compatability, but forgot to clear my "mongodb-binaries" cache (and still had the old version)

hasezoey:
pushed some commits to:
- add a `cleanup` script (removing all directories that made problems for me)
- removing `debug: true` from one test
- replaces all `/data` with `tmpDir.name` in `MongoInstance-test`
- adding `setGracefulCleanup` to resolve-config-test

nodkz:
`CURL_OPENSSL_3` means that required dependency for `mongod` binary not installed on your system.
We have the following message:
https://github.com/nodkz/mongodb-memory-server/blob/1b88c1c9eb57381bd9fa3050296db1a842264b1b/packages/mongodb-memory-server-core/src/util/MongoInstance.ts#L229-L232

Need to check which version of `CURL_OPENSSL` installed on your system and download the proper version of binaries. It's quite difficult to determine and provide a zero-config build-in solution.

hasezoey:
@nodkz like i said, i just had forgot to clear the mongodb binaries from the caches (had the versions before the mint integrations)

nodkz:
New version will be published in minutes.
If some issues occur let fix them in other PRs. 
For now, this PR became too long and how i feel it ready for publication.

Thanks a lot!!!

hasezoey:
> how i feel it ready for publication

like i said it was ready, today i just had done some cleanup and got the tests running

Thanks that i could help :)

hasezoey:
PS: i upgraded typegoose to the new version, the tests now work flawlessly on my system with memory enabled (and much faster) (still need to disabled in-memory, to debug what is stored in the db)

PR #261, opened 2020-01-09
Add TSDoc to MongoInstance

hasezoey:
- add TSDoc to MongoInstance
- add `@deprecated` to `getConnectionString` (TSDoc deprecated)
- rename some variables to make more sense
- fixed some types

(This is a draft, because of some `TODO: ask` notes)

### DIFF
--- a/packages/mongodb-memory-server-core/src/MongoMemoryServer.ts
+++ b/packages/mongodb-memory-server-core/src/MongoMemoryServer.ts
@@ -114,11 +114,7 @@ export default class MongoMemoryServer {
       })
       .catch((err) => {
         if (!this.opts.debug) {
-          // TODO: ask question
-          throw new Error(
-            `${err.message}\n\nUse debug option for more info: ` +
-              `new MongoMemoryServer({ debug: true })`
-          );
+          console.warn('Starting the instance failed, please enable "debug" for more infomation');
         }
         throw err;
       });
--- a/packages/mongodb-memory-server-core/src/MongoMemoryServer.ts
+++ b/packages/mongodb-memory-server-core/src/MongoMemoryServer.ts
@@ -16,6 +16,9 @@ import { isNullOrUndefined } from 'util';
 
 tmp.setGracefulCleanup();
 
+/**
+ * Starting Options
+ */
 export interface MongoMemoryServerOptsT {
   instance?: MongoMemoryInstancePropT;
   binary?: MongoBinaryOpts;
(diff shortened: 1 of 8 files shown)


packages/mongodb-memory-server-core/src/MongoMemoryServer.ts
```diff
@@ -108,6 +114,7 @@ export default class MongoMemoryServer {
       })
       .catch((err) => {
         if (!this.opts.debug) {
+          // TODO: ask question
           throw new Error(
             `${err.message}\n\nUse debug option for more info: ` +
```
hasezoey: should this error really be hidden behind debug?

packages/mongodb-memory-server-core/src/MongoMemoryServer.ts
```diff
@@ -127,16 +134,19 @@ export default class MongoMemoryServer {
    * @private
    */
   async _startUpInstance(): Promise<MongoInstanceDataT> {
-    const data: any = {};
+    /** Shortcut to this.opts.instance */
+    const instOpts = this.opts.instance ?? {};
     let tmpDir: DirResult;
+    const data: any = {
```
hasezoey: is this data object really wanted to be `any` instead of an own interface / properly typed?

packages/mongodb-memory-server-core/src/util/resolve-config.ts
```diff
@@ -15,7 +15,7 @@ export function setDefaultValue(key: string, value: string): void {
 
 let packageJson: Package | undefined;
 export function reInitializePackageJson(directory?: string): void {
-  packageJson = getPackageJson(directory || process.cwd());
```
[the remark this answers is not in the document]
hasezoey: thanks for pointing this out - i thought "just error(because of an empty string), instead of passing a default", changed in 11ff451330a2aeac45f2ff62a12cb440e5eb471e

packages/mongodb-memory-server-core/src/MongoMemoryServer.ts
hasezoey: changed in d448beadc5c48a3c8b1444f0e42bf208f410def4, is the message ok so?

hasezoey:
Note: changed 2 tests where `console.warn` gets called to mock the function and expect it to be called

packages/mongodb-memory-server-core/src/MongoMemoryServer.ts
hasezoey: changed in 70fbdabeb1a34153f59567619b0c5c09bb70a8d0, are you ok with the solution or did i miss something?

nodkz:
Let move this code to master.

Good job! 👍

Issue #263, opened 2020-01-15
@types/tmp/index"' has no default export


baptLesta:
Since version 6.2.2, there is a typescript issue when we build:
```
/node_modules/@types/tmp/index"' has no default export.

3 import tmp from 'tmp';
         ~~~


Found 1 error.
```

hasezoey:
nothing really changed, only an import for `tmp` got **removed** (because double), so, do you have any more information? did you clear your node_modules already?

hasezoey:
Addition: please add which package you are using

BigForNothing:
I just ran into a similar issue. 
```
> tsc
../../node_modules/mongodb-memory-server-core/lib/MongoMemoryServer.d.ts:3:17 - error TS7016: Could not find a declaration file for module 'tmp'. '../../node_modules/mongodb-memory-server-core/node_modules/tmp/lib/tmp.js' implicitly has an 'any' type.
  Try `npm install @types/tmp` if it exists or add a new declaration (.d.ts) file containing `declare module 'tmp';`

3 import tmp from 'tmp';
                  ~~~~~


Found 1 error.
```

Looking at the package.json for mongodb-memory-server-core, I see @types/tmp, but it is in devDependencies. Editing the installed package.json and moving `"@types/tmp": "0.1.0",` to dependencies will fix that.

Then I run into the issue above. There isn't a default export for @types/tmp, so changing the code to `import * as tmp from 'tmp'` should resolve it. The other alternative is consumers are required to have `"allowSyntheticDefaultImports": true` added to their tsconfig.json.

I would suggest not requiring the use of `allowSyntheticDefaultImports` along with moving any `@types` packages to dependencies if the lib is in dependencies. Both ensure consumers won't have to add any extra packages or change configs to use your package with TypeScript.

Recap:
1. move `@types/tmp` to dependencies
1. Change `import tmp from 'tmp'` to `import * as tmp from 'tmp'`

hasezoey:
>  but it is in devDependencies

this is by design, typescript type only packages should never be in anything other than devDependencies

BigForNothing:
That is not true at all actually. It was at one time the belief though. If you check out the npm page for [@types/npm](https://www.npmjs.com/package/@types/tmp) you will see the default instructions use `--save` and not `--save-dev` or yarn's `-D`.

[...]

PR #265, opened 2020-01-15
fix(mongodb-memory-server-core): change import syntax and moved packa…

BigForNothing:
…ges to dependencies (#263)

## Related Issues

- #1

### DIFF
--- a/packages/mongodb-memory-server-core/package.json
+++ b/packages/mongodb-memory-server-core/package.json
@@ -31,6 +31,9 @@
   },
   "homepage": "https://github.com/nodkz/mongodb-memory-server",
   "devDependencies": {
+    "rimraf": "^3.0.0"
+  },
+  "dependencies": {
     "@types/cross-spawn": "^6.0.1",
     "@types/decompress": "^4.2.3",
     "@types/dedent": "^0.7.0",
@@ -42,9 +45,6 @@
     "@types/mkdirp": "^0.5.2",
     "@types/tmp": "0.1.0",
     "@types/uuid": "3.4.6",
-    "rimraf": "^3.0.0"
-  },
-  "dependencies": {
     "camelcase": "^5.3.1",
     "cross-spawn": "^7.0.1",
     "debug": "^4.1.1",
--- a/packages/mongodb-memory-server-core/package.json
+++ b/packages/mongodb-memory-server-core/package.json
@@ -31,9 +31,6 @@
   },
   "homepage": "https://github.com/nodkz/mongodb-memory-server",
   "devDependencies": {
-    "rimraf": "^3.0.0"
-  },
-  "dependencies": {
     "@types/cross-spawn": "^6.0.1",
     "@types/decompress": "^4.2.3",
     "@types/dedent": "^0.7.0",
@@ -45,6 +42,9 @@
     "@types/mkdirp": "^0.5.2",
     "@types/tmp": "0.1.0",
     "@types/uuid": "3.4.6",
+    "rimraf": "^3.0.0"
+  },
+  "dependencies": {
     "camelcase": "^5.3.1",
     "cross-spawn": "^7.0.1",
     "debug": "^4.1.1",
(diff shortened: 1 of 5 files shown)


packages/mongodb-memory-server-core/src/MongoMemoryServer.ts
```diff
@@ -1,5 +1,5 @@
 import { ChildProcess } from 'child_process';
-import tmp from 'tmp';
+import * as tmp from 'tmp';
```
hasezoey: thanks didnt notice this when i was refactoring the file

packages/mongodb-memory-server-core/package.json
```diff
@@ -31,6 +31,9 @@
   },
   "homepage": "https://github.com/nodkz/mongodb-memory-server",
   "devDependencies": {
+    "rimraf": "^3.0.0"
+  },
+  "dependencies": {
```
hasezoey: like i said in #263, `@type` packages should never be in anything other than `devDependencies`

BigForNothing: Please see my comments on #263 as Microsoft's TypeScript handbook states you should be including them in dependencies.

hasezoey:
@BigForNothing ok, i understand now (sorry), but wouldnt it be better to put them into `peerDependencies`? because so it would not be an nested-install (like this package having its own version of some typings and some other package another version of typings)

BigForNothing:
I would say no, because you are still requiring people to install it. Also, peerDependencies was created to [solve](https://nodejs.org/es/blog/npm/peer-dependencies/) the plugins problem (think eslint plugins).

nodkz:
I always put `@types/*` packages to devDependencies. And I don't think that it's a good idea to install @types/* packages for a production build `npm install --production`. Anyway `mongodb-memory-server` package will be used as `devDependency` in most projects so there is no problem to put all `@types` packages to `dependencies`.

BUT damn, it's quite strange for me. I have a lot of other packages and had no problems.

So I asked Daniel (https://twitter.com/nodkz/status/1219160364014804992) about their recommendations in the handbook. I feel that they are incomplete or stale. Hope he answers soon.

nodkz:
Crap! I removed my question to Daniel.

@BigForNothing you provide incorrect information to us.
If you read handbook recommendations carefully - they are about how to prepare `@types` packages NOT regular packages 😈

Please move `@types` to `devDependencies` back 😉

So your PR should have just one change and it will be enough:
```diff
- import tmp from 'tmp';
+ import * as tmp from 'tmp';
```

hasezoey:
@nodkz just to note: i removed one import, the type for `DirResult` and replaced it with `tmp.DirResult` because tmp was already imported, and i didnt noticed that it was an *default* import and not what you mention now

nodkz:
@BigForNothing please check the new version of the package and confirm that error in your env is gone.

Tnx ;)

BigForNothing:
@nodkz I didn't actually. Did you go to the page that was linked from the issue thread? The example specifically shows a "regular" package with the name of "browserify-typescript-extension".

I think the confusion lies in that the very first section is about publishing declaration files.

Here is a [thread](https://github.com/microsoft/types-publisher/issues/81) where Daniel talks about this. While it's for MS's type publisher tool, it's in reference to why "--save" started showing up as the recommended install method instead of "--save-dev".

> Using --save-dev is fine when you're writing a simple application, and it won't be used as a library. The problem comes along when you might have dependencies. If you stored your type declarations in your devDependencies, your consumers would not automatically get your type declarations. They'd then have to manually add your declaration file dependencies as their devDependencies.

> Given that breaking consumers is a worse problem than slightly-larger packages, we've made --save the default in our documentation.

And im a reply further down: 

> My thought is dependencies since you're written as a library with dependents.

You can see that the intention would apply to this project, and "@types/tmp" should be in dependencies.

Im on mobile at the moment, so I'll test the latest changes when I'm on the computer.

PR #266, opened 2020-01-22
Add @types/tmp to dependencies

ronjouch:
Not including it in devDeps means that `strict: true` TSC projects
using mongoms currently fail to compile with:

```
node_modules/mongodb-memory-server-core/lib/MongoMemoryServer.d.ts:3:22 - error TS7016:
  Could not find a declaration file for module 'tmp'.
 'node_modules/mongodb-memory-server-core/node_modules/tmp/lib/tmp.js' implicitly has an 'any' type.
  Try `npm install @types/tmp` if it exists or add a new declaration (.d.ts) file containing `declare module 'tmp';`

3 import * as tmp from 'tmp';
                       ~~~~~
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
     "@types/jest": "24.0.24",
     "@types/mongodb": "3.3.13",
     "@types/node": "12.12.21",
-    "@types/tmp": "0.1.0",
     "@typescript-eslint/eslint-plugin": "2.12.0",
     "@typescript-eslint/parser": "2.12.0",
     "cross-env": "^6.0.3",
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "@types/jest": "24.0.24",
     "@types/mongodb": "3.3.13",
     "@types/node": "12.12.21",
+    "@types/tmp": "0.1.0",
     "@typescript-eslint/eslint-plugin": "2.12.0",
     "@typescript-eslint/parser": "2.12.0",
     "cross-env": "^6.0.3",
--- a/packages/mongodb-memory-server-core/package.json
+++ b/packages/mongodb-memory-server-core/package.json
@@ -31,6 +31,9 @@
   },
   "homepage": "https://github.com/nodkz/mongodb-memory-server",
   "devDependencies": {
+    "rimraf": "^3.0.0"
+  },
+  "dependencies": {
     "@types/cross-spawn": "^6.0.1",
     "@types/decompress": "^4.2.3",
     "@types/dedent": "^0.7.0",
@@ -42,9 +45,6 @@
     "@types/mkdirp": "^0.5.2",
     "@types/tmp": "0.1.0",
     "@types/uuid": "3.4.6",
-    "rimraf": "^3.0.0"
-  },
-  "dependencies": {
     "camelcase": "^5.3.1",
     "cross-spawn": "^7.0.1",
     "debug": "^4.1.1",


BigForNothing:
@ronjouch see #265 

It should be in dependencies btw, not devDependencies as I discussed in that PR.

ronjouch:
> see #265. It should be in dependencies btw, not devDependencies as I discussed in that PR.

@BigForNothing I think you're right, as it's used by library code exposed to consumers like my project.

And oh, now I understand: I edited the wrong file; this is a monorepo and I should touch code in `packages`. Updating the PR, thanks for the poke.

ronjouch:
@hasezoey @nodkz I pushed a new fix; ready for review.

Coming back to discussion in #265, I think @BigForNothing is right and @nodkz is missing the point. @nodkz , you say:

> I always put `@types/*` packages to devDependencies. And I don't think that it's a good idea to install @types/* packages for a production build `npm install --production`. Anyway `mongodb-memory-server` package will be used as `devDependency` in most projects so there is no problem to put all `@types` packages to `dependencies`.

To the best of my current knowledge, this is incorrect. The practical meaning of `devDependencies` is that they're not going to be installed if the packaged is installed transitively (as opposed to `npm install`ing it). That's my case, I'm using mongoms in my project. So, by putting the types in `devDeps`, they're _not_ going to be installed in my `strict: true` TS project depending on mongoms, and compiling the tests (using mongoms) will fail. Currently, by chance, `@types/tmp` is the only missing package in my `node_modules`, but all your types for prod/lib code belongs to `dependencies`, not `devDependencies`.

I stumbled on the same problem at work, and my policy is now:

- types for *prod* code in deps, along with their respective package
- types for *test* code in devDeps, along with their respective package.

For example, here's the project I'm working on; see how prod types are with prod libs in deps, and same for test types with test libs:

```json
  "devDependencies": {
    "@types/jest": "24.x",
    "@types/supertest": "2.x",
    "@types/tmp": "0.x",
    "@typescript-eslint/eslint-plugin": "2.x",
    "@typescript-eslint/parser": "2.x",
    "eslint": "6.x",
    "eslint-config-prettier": "6.x",
    "eslint-plugin-prettier": "3.x",
    "jest": "25.x",
    "jest-junit": "10.x",
    "mongodb-memory-server": "6.x",
    "npm-audit-resolver": "2.x",
    "prettier": "1.x",
    "supertest": "4.x",
    "typescript": "3.x"
  },
  "dependencies": {
    "@types/body-parser": "1.x",
    "@types/express": "4.x",
    "@types/mongoose": "5.x",
    "@types/node": "10.x",
    "@types/request-promise": "4.x",
    "@unito-private/babylon-client": "0.x",
    "@unito-private/ucommon": "3.x",
    "body-parser": "1.x",
    "express": "4.x",
    "mongoose": "5.x",
    "request-promise": "4.x",
    "source-map-support": "0.x"
  }
```

Kinda sure about all that, but I welcome Daniel's take, maybe I'm missing something 🙂.

hasezoey:
@ronjouch i would like to not be involved in this discussion *for this project*, because i dont know what would be the right solution, and in its current state is the only thing i know and always done like it

anyway, my take was always something like "if it isnt included (/needed) after the compilation, dont put it in normal dependencies" (and it always worked, but i didnt have many projects in strict mode, because of the nature of these projects)

ronjouch:
> I would like to not be involved in this discussion _for this project_, because i dont know what would be the right solution, and in its current state is the only thing i know and always done like it

@hasezoey okay, leaving you out of it, and continuing the chat with @nodkz.

@nodkz , what I'm suggesting is that I did fiddle with that quite a bit at work, and I think I know the correct solution to satisfy `strict: true` projects, and I don't see a reason not to support those.

> anyway, my take was always something like "if it isnt included (/needed) after the compilation, dont put it in normal dependencies" (and it always worked, but i didnt have many projects in strict mode, because of the nature of these projects)

Correct! The gotcha here is that, for `strict: true` TS projects, types **are needed** after compilation. So, as you say, they should be put in "normal dependencies".

Now, one may argue they're the exception rather than the rule. As a happy TS user, I'm biased to disagree, because I wouldn't want to add to my package.json the types needed to compile dependencies X, Y, Z. To me, types belongs to the build artifact packaged to library users, much like source maps.

BigForNothing:
@ronjouch Well said! I could be wrong, but I believe --isolatedModules require it too. 

All of our projects at work require the strict: true setting, and I've never had a personal project without it.

lirbank:
@types/tmp should be in `dependencies` (not `devDependencies`). See:
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies

nodkz:
@ronjouch thanks for PR and the detailed explanation of `strict` mode in TS.
Merged! 👍

PR #270, opened 2020-02-26
Logging rewrite

### LINKED ISSUE #262: Replace custom debug functions with logging module
*from #214*

i would suggest to replace all `this.debug` calls with some logging module, like `debug` or `loglevel`

---

Pros for loglevel:
- has many levels that are easy to access `logger.<level>`

Cons for loglevel:
- cant be activated via environment (needs to be activated with `setLogLevel`)
- cant be activated for one specific file

---

Pros for debug:
- can be activated via environment
- can be activated for only one file / can be filtered for one file

Cons for debug:
- needs to set a "new logger" in every file
- dosnt have any levels

hasezoey:
- Add Dev-dependency "@types/debug" (the normal `debug` package is already an dependency, but never used)
- Change Typescript lib from "es2017" to "esnext" (just to be sure to be able to work with new things, they will get compiled down)
- Added some Logging to "MongoBinaryDownloadUrl"
- Transformed logging to use "debug"
- "resolve-config" now gets loaded in index as an side-effect (to load MONGOMS_DEBUG)
- Added "Enable Debug Mode" section to README
- Reorganized the README's documented options
- Changed `MONGOMS_DEBUG` to enable `debug.('MongoMS:*')`
- `DEBUG=MongoMS:*` will now work as an environment variable too (cant be disabled)

The only file where i didnt remove `this.debug` is `MongoInstance` because it contains the STDOUT of Mong-Instances

## Related Issues

- fixes #262 

---

please know that this is not intended to be merged right away

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ _Note: the package does try to download `mongod` upon server start if it cannot 
 
 ## Usage
 
-### Simple server start:
+### Simple server start
 
 ```js
 import { MongoMemoryServer } from 'mongodb-memory-server';
@@ -108,7 +108,7 @@ mongod.getInstanceInfo();
 // special childProcess killer will shutdown it for you
 ```
 
-### Available options
+### Available options for MongoMemoryServer
 
 All options are optional.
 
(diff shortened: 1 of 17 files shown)


nodkz:
Looks good👍
Is it done? I'm ready to merge it!

hasezoey:
@nodkz yes it is done, just wanted to get some feedback and see if i have overseen something - and it was in the state of `not really necessary`, so just thought "put it in an draft to not accidentally merge it" (and because of the decreased coverage)

PS: just noticed an flaw in `MongoMemoryServer-test` `should throw an error if _startUpInstance throws an unknown error` because it expects `console.warn` to have been called, which it dosnt when debug is activated for that file
-> fixed it in f988f0ec2d862ed05e1358960c372399c1cd9455

nodkz:
I looked from mobile and all looks good. But maybe i miss something. If so - there is no problem to fix it in the next versions. 😉

Decreased coverage is not a problem. For newcomers Codecov helps to not forget about tests. And also it sends nice messages with chart 😜

nodkz:
Great PR and description! 👌
Tnx 🙏

PR #272, opened 2020-02-28
Small Changes

hasezoey:
- Change watch script to use "--watchAll"
- Remove "this.admin" because it is not used outside of one function
- Add an counter to how many instances are spawned for better logging


nodkz:
👍

Issue #278, opened 2020-03-10
restarting replset with already existing files fails with "already initialized"


sakulstra:
Disclaimer: I know that not the intention of this library at first place and I probably shouldn't abuse it this way.

## Versions

- NodeJS: 12.x
- mongodb-memory-server-*: 6.3.1
- mongodb: default that comes for arch
- system: arch

package: mongo-memory-server

## What is your question?

The issue i'm trying to solve is that within out company there are people accross all OSes which want to play around with our not yet released api which relies on a mongodb running on the system.

For our stuff to work we need a somehwat up-to-date version of mongodb with a replicaset, so we created a simple script which uses `'mongodb-memory-server` to start a mongodb without the setup hassle and it works great :+1:

One problem though is that in some cases we'd like to persist the test data trough system reboots and stuff like that :sweat_smile: 

So I was wondering if there is a way to start mongodb with previously persisted data similar to `mongod --dbpath database --replSet rs`? I saw there's a:
```
instanceOpts: [
    {
      port: 27017,
      dbPath: "database",
    },
  ],
```
option to specify the path, but when reusing a path from a previous run I get the following error
```
(node:29641) UnhandledPromiseRejectionWarning: MongoError: already initialized
    at MessageStream.messageHandler (/home/lukass/Documents/sputnik/node_modules/mongodb/lib/cmap/connection.js:261:20)
    at MessageStream.emit (events.js:311:20)
    at processIncomingData (/home/lukass/Documents/sputnik/node_modules/mongodb/lib/cmap/message_stream.js:144:12)
    at MessageStream._write (/home/lukass/Documents/sputnik/node_modules/mongodb/lib/cmap/message_stream.js:42:5)
    at doWrite (_stream_writable.js:441:12)
    at writeOrBuffer (_stream_writable.js:425:5)
    at MessageStream.Writable.write (_stream_writable.js:316:11)
    at Socket.ondata (_stream_readable.js:714:22)
    at Socket.emit (events.js:311:20)
    at addChunk (_stream_readable.js:294:12)
```

hasezoey:
dbPath should be an absolute path to avoid conflictions

if you give an dbPath, it will be used across all instances, this is why when more than one are started, it tries to use the same path, what is already locked by another instance

Sorry, but currently the only way to use replset with more than one instance, is to not specify an dbPath to automatically create temporary directories (in `/tmp`)

sakulstra:
[...]

PR #283, opened 2020-03-13
fix: catch reinit of replicaset

boredland:
- #278

### DIFF
--- a/packages/mongodb-memory-server-core/src/__tests__/replset-single-restart-test.ts
+++ b/packages/mongodb-memory-server-core/src/__tests__/replset-single-restart-test.ts
@@ -1,8 +1,6 @@
 import MongoMemoryReplSet from '../MongoMemoryReplSet';
 import * as tmp from 'tmp';
 
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 600000;
-
 let tmpDir: tmp.DirResult;
 beforeEach(() => {
   tmpDir = tmp.dirSync({ prefix: 'reuse-mongo-mem-', unsafeCleanup: true });
@@ -43,5 +41,5 @@ describe('single-member replica set', () => {
     const replSetAfter = new MongoMemoryReplSet(opts);
     await replSetAfter.waitUntilRunning();
     await replSetAfter.stop();
-  });
+  }, 600000);
 });
--- a/packages/mongodb-memory-server-core/src/__tests__/replset-single-restart-test.ts
+++ b/packages/mongodb-memory-server-core/src/__tests__/replset-single-restart-test.ts
@@ -1,0 +1,47 @@
+import MongoMemoryReplSet from '../MongoMemoryReplSet';
+import * as tmp from 'tmp';
+
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 600000;
+
+let tmpDir: tmp.DirResult;
+beforeEach(() => {
+  tmpDir = tmp.dirSync({ prefix: 'reuse-mongo-mem-', unsafeCleanup: true });
+});
+
+afterEach(() => {
+  tmpDir.removeCallback();
+});
+
+const sleep = (ms: number) => {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+};
+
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -46,3 +46,6 @@ packages/mongodb-memory-server-core/lib
 coverage
 .nyc_output
 package-lock.json
+
+# testfiles
+reuse*
```
hasezoey: why do these need to be ignored?

packages/mongodb-memory-server-core/src/__tests__/replset-single-restart-test.ts
```diff
@@ -0,0 +1,32 @@
+import MongoMemoryReplSet from '../MongoMemoryReplSet';
+import fs from 'fs';
+import rimraf from 'rimraf';
+
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 600000;
+
+describe('multi-member replica set', () => {
+  it('should start multiple times', async () => {
+    const testDir = fs.mkdtempSync('reuse_');
```
hasezoey: please refactor to use tmp (package)

yarn.lock
```diff
@@ -1839,6 +1839,14 @@
   resolved "https://registry.yarnpkg.com/@types/node/-/node-12.12.21.tgz#aa44a6363291c7037111c47e4661ad210aded23f"
   integrity sha512-8sRGhbpU+ck1n0PGAUgVrWrWdjSW2aqNeyC15W88GRsMpSwzv6RJGlLhE7s2RhVSOdyDmxbqlWSeThq4/7xqlA==
 
+"@types/rimraf@^2.0.3":
```
hasezoey: i think this should be removed here too

hasezoey:
are you sure you reverted the lockfile correctly?

boredland:
> are you sure you reverted the lockfile correctly?
> -> seems like more is now removed than earlier commits added

I"m sure I reverted that one correctly, but now I just squashed those changes together to cover the unnecessary changes.

hasezoey:
@boredland checked it out locally, and the yarn.lock is still not like the master one (many things removed and set to old versions)

to fix this run `yarn install` in the root of the repo or in the directory the yarn.lock file is in and commit it, you will see if you compare that with nodkz/master that only 3 package versions have changed
-> i would have pushed it to your branch, but it seems like "Allow edits from maintainers" is disabled for this pr / is not possible when you dont own the repo/branch

packages/mongodb-memory-server-core/src/__tests__/replset-single-restart-test.ts
```diff
@@ -0,0 +1,47 @@
...
+afterEach(() => {
+  tmpDir.removeCallback();
+});
+
+const sleep = (ms: number) => {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+};
+
+describe('single-member replica set', () => {
+  it('should start multiple times', async () => {
```
hasezoey: maybe an special test-timeout should be set for this test

packages/mongodb-memory-server-core/src/__tests__/replset-single-restart-test.ts
```diff
@@ -0,0 +1,47 @@
+import MongoMemoryReplSet from '../MongoMemoryReplSet';
+import * as tmp from 'tmp';
+
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 600000;
```
hasezoey: please remove this, otherwise it might affect other tests too

sakulstra:
hayho, what's the status here?

@hasezoey can we somehow help pushing this forward? :pray:

nodkz:
Everything is ok! Bump it!

Tnx 🙏

Ranguna:
Has this been released already ?
I can't seem to find these changes on my locally installed `mongodb-memory-server-core`.

hasezoey:
@Ranguna yes since 6.4.0

Ranguna:
@hasezoey all right, thanks!

PR #288, opened 2020-03-23
Add support for rhel & coreos 8

GlenTiki:
This is a quick one, I've just upgraded my servers to coreos 8 and this is causing issues.


nodkz:
Thanks!

PR #293, opened 2020-04-02
change "waitForPrimary" regex case-insensitive

### LINKED ISSUE #292: MongoMemoryReplSet hangs in docker container for mongo 4.4
## Versions

- NodeJS: 12.16.1
- mongodb-memory-server-*: 6.5.1
- mongodb: 4.4-latest
- system: buster-slim docker image

package: mongo-memory-server

## What is the Problem?

Similar to #289 but with `MongoMemoryReplSet`

## Code Example

```ts
const mongodb = require('mongodb-memory-server');
(async () => {
    const replSet = new mongodb.MongoMemoryReplSet({
      replSet: { storageEngine: 'wiredTiger' },
      debug: true,
    });
    mongoUri = await replSet.getUri();
    await replSet.stop();
})();
```

I think the message return by mongo server on stdout is : 
`Transition to primary complete; database writes are now permitted`
But the lib check :
`transition to primary complete; database writes are now permitted`
The first char is capitalized in mongo 4.4

https://github.com/nodkz/mongodb-memory-server/blob/master/packages/mongodb-memory-server-core/src/util/MongoInstance.ts#L274

hasezoey:
change "waitForPrimary" regex case to case-insensitive

## Related Issues

- fixes #292

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
@@ -271,7 +271,7 @@ export default class MongoInstance {
       }
     } else if (/\*\*\*aborting after/i.test(line)) {
       this.instanceFailed('Mongod internal error');
-    } else if (/transition to primary complete; database writes are now permitted/.test(line)) {
+    } else if (/transition to primary complete; database writes are now permitted/i.test(line)) {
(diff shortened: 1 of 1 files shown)


nodkz:
Merged via https://github.com/nodkz/mongodb-memory-server/pull/294

PR #294, opened 2020-04-02
insensitive the regexp when a primary is elected

### LINKED ISSUE #292: MongoMemoryReplSet hangs in docker container for mongo 4.4

lafeuil:
insensitive the regexp when a primary is elected because on mongo 4.4 the stdout message has a capital letter

## Related Issues
fixes #292

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
@@ -271,7 +271,7 @@ export default class MongoInstance {
       }
     } else if (/\*\*\*aborting after/i.test(line)) {
       this.instanceFailed('Mongod internal error');
-    } else if (/transition to primary complete; database writes are now permitted/.test(line)) {
+    } else if (/transition to primary complete; database writes are now permitted/i.test(line)) {
       this.isInstancePrimary = true;
       this.debug('Calling all waitForPrimary resolve functions');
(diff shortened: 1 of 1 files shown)


hasezoey:
i guess we both made an pr (#293, #294)

nodkz:
@lafeuil Thanks a lot! 👍

PS. @hasezoey thanks for your PR. 🙏 Anyway, we should salute new contributors too. So I choose this PR ;)  Closes #293

hasezoey:
@nodkz definitely understand, would have chosen this way too :+1:

PR #310, opened 2020-05-19
Fix #308 - read strict-ssl from npm config

### LINKED ISSUE #138: Not work on Windows 10 Home edition
Hi,
I'm using jest for testing.
The issue found on laptop that runs win 10 home.
However everything works fine under Manjaro and Bash on Ubuntu on Windows, on Win 10 Pro works great too.
I have tried a different versions of node (11.8 - 11.10, 10.15.1 which is LTS).
Problem is infinite download of MongoDB, in normal case It's downloading while npm install in process.
I have tried setting jest.setTimeout to 10e5 which is pretty much, once download finishes another one starts.

### LINKED ISSUE #308: npm strict-ssl checking doesn't work
## Versions

- NodeJS: v12.16.3
- mongodb-memory-server-*: 6.6.0
- system: MacOS
- npm: 6.14.4

package: mongo-memory-server

## What is the Problem?

I was hitting this issue: https://github.com/nodkz/mongodb-memory-server/issues/162 and saw that it should have been fixed in V6.6.0. But the fix to specify `strict-ssl=false` doesn't work for me on MacOS. 

## Do you know *why* it happens?

After a bit of digging the reasons seems to be (at least on MacOS):

- The environment variable at runtime is named `npm_config_strict_ssl` not `npm_config_strict-ssl` as in the code.
- When you specify `--strict-ssl=false` npm does not populate the variable as the string 'false'. It is an empty string. If you specify as true or omit it is set to 'true'.

NOTE: This appears like it could affect other places that pickup `npm_config_*` variables. e.g. `npm_config_https-proxy`

mlev:
Fix for #308 :

- Read from environment variable `npm_config_strict_ssl` (i.e. all underscores)
- Will honour whatever the npm config value is. npm/yarn will set true properties to the string 'true' and false properties to an empty string. The current npm default is `true` so even without specifying `--strict-ssl` this will be set to 'true' in environment variables.

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
@@ -146,7 +146,7 @@ export default class MongoBinaryDownload {
       process.env.HTTPS_PROXY ||
       process.env.HTTP_PROXY;
 
-    const strictSsl = process.env['npm_config_strict-ssl'] === 'false' ? false : true;
+    const strictSsl = process.env.npm_config_strict_ssl === 'true';
 
     const urlObject = url.parse(downloadUrl);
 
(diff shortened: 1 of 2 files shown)


nodkz:
Thanks!

Issue #226, opened 2019-10-03
Infinite blockage when version cant be downloaded


dcurletti:

AJRdev:
@dcurletti Normally we already handle MongoDB v4.2 in this package, do you experience any issues with this particular version ?
If not this issue will be closed 🙂

duro:
@AJRdev I am also experiencing issues getting the package to work with Mongo 4.2. The issue I am having is that when I specify that I want 4.2 like so in the setup:

```
mongoServer = new MongoMemoryReplSet({
  binary: {
    version: '4.2.0'
  },
  replSet: { storageEngine: 'wiredTiger' },
})
```

This works just fine locally on my Mac, however when this is running in CircleCI I get the following error:

```
   console.warn ../../node_modules/mongodb-memory-server-core/lib/util/MongoBinaryDownloadUrl.js:186
     Unknown linux distro Raspbian, falling back to legacy MongoDB build
 Downloading MongoDB 4.2.0: NaN % (0mb / NaNmb)
   console.error ../../node_modules/jest-jasmine2/build/jasmine/Env.js:289
     Unhandled error
   console.error ../../node_modules/jest-jasmine2/build/jasmine/Env.js:290
     Error: Too small (243 bytes) mongod binary downloaded from https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-4.2.0.tgz
```

The very odd thing is, I am running this in a monorepo which has many other services being tested that are using `mongodb-memory-server`, but are still just using the default version of `4.0.3` and even on CircleCI, those work just fine.

It is only the service trying to use 4.2 which fails on Circle.

The other using 4.0.3 still throw the warning: `Unknown linux distro Raspbian, falling back to legacy MongoDB build`, but the successfully download that fallback version just fine.

kunalbhatt:
> @AJRdev I am also experiencing issues getting the package to work with Mongo 4.2. The issue I am having is that when I specify that I want 4.2 like so in the setup:
> 
> ```
> mongoServer = new MongoMemoryReplSet({
>   binary: {
>     version: '4.2.0'
>   },
>   replSet: { storageEngine: 'wiredTiger' },
> })
> ```
> 
> This works just fine locally on my Mac, however when this is running in CircleCI I get the following error:
> 
> ```
>    console.warn ../../node_modules/mongodb-memory-server-core/lib/util/MongoBinaryDownloadUrl.js:186
[...]

PR #327, opened 2020-07-27
add debug log for url

hasezoey:
- fix(MongoBinaryDownload): add debug log for url
- fix(MongoBinaryDownload): put "stdout.write" in an "isTTY" and otherwise use "console.log"
- rename script `cleanup` to `clean`

## Related Issues

- suggested in #226


nodkz:
Good thing! Thanks!

Issue #298, opened 2020-04-06
@types/get-port@4.2.0 dependency is deprecated


lafeuil:
## Versions

- NodeJS: 12
- mongodb-memory-server-*: 6.5.2
- mongodb: 4.2-next
- system: Linux

package: mongo-memory-server

## What is the Problem?
After installing the mongo-memory-server, I have this deprecated warning :  
```
npm WARN deprecated @types/get-port@4.2.0: This is a stub types definition. get-port provides its own type definitions, so you do not need this installed.
```

This dependency is in mongodb-memory-server-core package.
```
$ npm ls @types/get-port
[...]
└─┬ mongodb-memory-server@6.5.2
  └─┬ mongodb-memory-server-core@6.5.2
    └── @types/get-port@4.2.0 
```

## Do you know *why* it happenes?

*no*

lafeuil:
I don't check but I think you can remove this dependency because the types definition are defined in the main package `get-port`.

hasezoey:
@lafeuil it will get removed once we move from `get-port` to `portfinder` #287

PR #328, opened 2020-07-27
fix: remove stub package "@types/get-port"

hasezoey:
## Related Issues

- closes nodkz/mongodb-memory-server#298


nodkz:
I'm glad to see that more and more packages provide TS definitions out of the box.
Thanks for PR!

PR #342, opened 2020-08-24
Match README description for checkMD5

jpumford:
This PR updates the code to follow the README.

I noticed the checkMD5 option was not available on the `MongoMemoryServer` class constructor, though the README indicated it is an option.

Here's a quick PR passing through the option to do an MD5 check for the downloaded mongo binary.

I also added two quick log statements in the MD5 check function. If that's a problem I can take them out, but they were useful to me and may be useful to someone else running in debug mode.


nodkz:
Thanks for your contribution!

PR #353, opened 2020-09-08
fix(package): move non-exposed @types to devDependencies

### LINKED ISSUE #286: Unnecessary non-devDependencies in mongodb-memory-server-core
## Versions

- mongodb-memory-server-*: 6.3.3

package: mongodb-memory-server-core

## What is the Problem?

`mongodb-memory-server-core` installs a lot of unnecessary dependencies (e.g. all the `@types/...`) in a consuming project: 

https://github.com/nodkz/mongodb-memory-server/blob/52289e85710ac0b552ba50e9c9db32d4aef26547/packages/mongodb-memory-server-core/package.json#L49-L60

Please move them into `devDependencies`.

hasezoey:
- move all non-exposed types from `@types` to `devDependencies`

## Related Issues

- fixes #286


nodkz:
Yeah, at last addition of @types/* to dependencies became clearer – must be included only if they exposed.
When we discussed a month or two ago I didn't understand ts recommendations and was feeling it awkward.

Thanks!

PR #354, opened 2020-09-08
add workflow "stale"

hasezoey:
- add github workflow "stale"

settings:
- runs once every day
- runs only on these tags: `question,can't reproduce,Needs More Information,needs repro script,stale`
- definitely doesnt run on these tags: `help wanted,docs,enhancement,feature,dependencies`
- marks issues as stale if there is no activity in 60 days
- closes issues when the date tagging with `stale` is older than 7 days
- removes stale tag on activity


nodkz:
Thanks a lot! Amazing tool!

PR #360, opened 2020-09-09
fix(MongoBinaryDownload): print used URL in 404 Error

hasezoey:
- print the used URL in the Error as well, because issues mostly get reported without debug enabled


nodkz:
Ouch, too fast.
We said about the problem. But do not provide any advice. 

Something like: 
"
Please, try to use different version `new MongoMemoryServer({ binary: { version: 'X.Y.Z' } })`. List of available versions can be found here: 
  https://www.mongodb.org/dl/linux for Ubuntu, RHEL, Debian, SUSE, Amazon
  https://www.mongodb.org/dl/osx for OSX
  https://www.mongodb.org/dl/win32 for Windows
"

hasezoey:
@nodkz will include in #363

PR #358, opened 2020-09-08
Upgrade build and publish procedures 

nodkz:
- Add new packages `mongodb-memory-server-global-4.2`, `mongodb-memory-server-global-4.4`
- Migrate from `circle ci` to `github actions`. If we started using them so better to unify CI system.
- Migrate from `lerna publish` back on `semantic-release`. I found a solution in its configs on how to support monorepos with multiple packages. I cannot live without semantic-release comments in PR & issues.

PR #367, opened 2020-09-10
build: remove packages below 4.2 from release

hasezoey:
- remove "global-x.x" packages below 4.2 from releases
- README: add deprecation notice

---

This PR removes the packages 3.4, 3.6 and 4.0 from the automated release,
after this pr has been merged the following commands should be run (by @nodkz)
```sh
npm deprecate mongodb-memory-server-global-3.4 "Deprecated, please use an version >= 4.2 (mongodb-memory-server-global-4.2)"
npm deprecate mongodb-memory-server-global-4.0 "Deprecated, please use an version >= 4.2 (mongodb-memory-server-global-4.2)"
```
the commands above will deprecated the entire package (3.4, 3.6, 4.0) with the message provided, this message will show on install and on the npm pages

---

let me know if an message should be updated

PS: if you need an example of a deprecated package, [i once deprecated one](https://www.npmjs.com/package/@hasezoey/typegoose)

### DIFF
--- a/.releaserc.js
+++ b/.releaserc.js
@@ -50,6 +50,13 @@ module.exports = {
       '@semantic-release/npm',
       {
         npmPublish: true,
+        pkgRoot: './packages/mongodb-memory-server-global-3.6',
+      },
+    ],
+    [
+      '@semantic-release/npm',
+      {
+        npmPublish: true,
         pkgRoot: './packages/mongodb-memory-server-global-4.2',
       },
     ],
--- a/.releaserc.js
+++ b/.releaserc.js
@@ -50,27 +50,6 @@ module.exports = {
       '@semantic-release/npm',
       {
         npmPublish: true,
-        pkgRoot: './packages/mongodb-memory-server-global-3.4',
-      },
-    ],
-    [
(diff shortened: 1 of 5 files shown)


nodkz:
How I remember we wanted to deprecate only
- 4.0 – too much missing distros for different os
- 3.4 – it's quite old and not widely  used

But 3.6 is still quite popular.

nodkz:
So maybe we keep all these packages?
I don't think that their removal helps us somehow.

hasezoey:
> How I remember we wanted to deprecate only

[in this comment](https://github.com/nodkz/mongodb-memory-server/pull/358#issuecomment-688882388) i said `maybe to deprecate the packages <4.2`

> But 3.6 is still quite popular.

ok, then i would remove the notes for this

hasezoey:
> So maybe we keep all these packages?
> I don't think that their removal helps us somehow.

this is just an proposal, the mentioned versions are just not available for current distros anymore - so i just thought that these version should be deprecated in favor of newer ones

nodkz:
Deprecated

Screen Shot 2020-09-17 at 01 49 26

PR #368, opened 2020-09-10
fix(MongoMemoryReplSet): comment-out older hack

### LINKED ISSUE #366: MongoMemoryReplSet: _initReplSet: is "shouldCheckForSessionSupport" still needed?
is the following line still needed? i tested without it, and no errors

https://github.com/nodkz/mongodb-memory-server/blob/6425c78edb7f94187bc12deda21927a4efabad70/packages/mongodb-memory-server-core/src/MongoMemoryReplSet.ts#L254-L262

and the mentioned code lines changed too, to 

https://github.com/mongodb/node-mongodb-native/blob/c16ec43a3cd9a751c11b10b1b6b2bc3cc4d933c4/src/operations/execute_operation.ts#L67-L92

hasezoey:
- comment out the hack added by 470f09485d5caa9062fc1c24fb5efc0eb3d7d32f
(almost reverts 470f09485d5caa9062fc1c24fb5efc0eb3d7d32f)

## Related Issues

- closes #366

### DIFF
--- a/packages/mongodb-memory-server-core/src/MongoMemoryReplSet.ts
+++ b/packages/mongodb-memory-server-core/src/MongoMemoryReplSet.ts
@@ -253,13 +253,7 @@ export default class MongoMemoryReplSet extends EventEmitter {
 
       // MongoClient HACK which helps to avoid the following error:
       //   "RangeError: Maximum call stack size exceeded"
-      //   in ./node_modules/mongodb/lib/core/sdam/topology_description.js:263:51
-      // Under the hood initiated server topology has type `RSGhost`
-      //   and this server does not support session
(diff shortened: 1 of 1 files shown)


packages/mongodb-memory-server-core/src/MongoMemoryReplSet.ts
```diff
@@ -253,13 +253,7 @@ export default class MongoMemoryReplSet extends EventEmitter {
 
       // MongoClient HACK which helps to avoid the following error:
```
hasezoey, line 254: i still let a bit of the old hack-comment to know what this line was for and for what issues to look out for

PR #377, opened 2020-09-13
fix(resolve-config): improve code

### LINKED ISSUE #187: Options can only be set in outermost package.json
The README says that options can be [set in package.json](https://github.com/nodkz/mongodb-memory-server#options-which-can-be-set-via-packagejsons-config-section). Given the following project structure, the options can only be set in `project/package.json`, and not in `project/subproject/package.json`. (Any options set in `project/subproject/package.json` are ignored.):

    project
    |-- subproject
    |   |-- package.json
    |   +-- ...
    |-- package.json
    +-- ...

This seems counter-intuitive to me. I'd suggest that options set in the subproject should at least have higher precedence than options in the project.

The cause of this is the loop in [`getPackageJson()`](https://github.com/nodkz/mongodb-memory-server/blob/master/packages/mongodb-memory-server-core/src/util/resolve-config.ts#L11) which exhausts the find-package-json iterator, returning the last value rather than the first. Is there a specific need to use the outermost package.json? If not then I'd suggest just taking the first value and leaving it at that. If we do need values from the outermost package.json then I'd suggest that `resolveConfig()` could reconcile the values extracted from each package.json, giving priority to the subprojects first.

Thoughts? I'm happy to make a pull request.

hasezoey:
fix(resolve-config): improve code
- envToBool: default to empty string (improve coverage)
- combine functions "getPackageJson" and "reInitializePackageJson"
- rename function "reInitializePackageJson" to "findPackageJson" (backwards compatible)
- findPackageJson: add tsdoc
- setDefaultValue: add tsdoc
- findPackageJson: log what package.json (path) is used
- rename variable "packageJson" to "packageJsonConfig"
- packageJsonConfig: default to empty object
- other files: apply "envToBool" defaulting
- resolve-config-test: apply function name change

style(README): add note about what package.json is used
- add an additional note about what package.json is used by default
- add example on how to change the package.json used

## Related issues

- i think this also fixes #187 (this pr adds logging about which to use, and examples on how to change it)


nodkz:
Merge rain for today 🤣

Screen Shot 2020-09-17 at 01 50 38

Every PR had some conflict with yarn.lock, readme TOC, package.json. So we agreed to avoid merge commits but for special days (like today 🤣) let we close eyes.

PR #383, opened 2020-09-19
feat: skip binary download when binary tar exists

### LINKED ISSUE #303: Reusing already dowloaded binary tar.
## Versions
- NodeJS: 13.7.0
- mongodb-memory-server-core: 6.5.2
- mongodb: any
- system: any

package: mongo-memory-server-core

## What is your question?
Does it make sense to have a check for already downloaded mongod binary tar, so to skip the download process again and continue with the unzipping process? 

### Reason
On slow networks, it may happen that binary got downloaded(after jest timeout) but due to jest timeout settings test cases failed. In this case, running test cases again will get the already downloaded binary tar(after performing integrity test) and use that to run the mongo instance.

as-ajitsingh:
- Added check to skip binary tar download, if the required tar is already present in the location

## Related Issues
- closes #303


hasezoey (approved):

nodkz:
Thanks a lot 👍

PR #386, opened 2020-09-24
explicitly export default from mms

murpjack:
This line explicitly exports `exports default` - mongoMemoryServer - received from require.
`if (p === "default") exports.default = m[p];`

### DIFF
--- a/packages/mongodb-memory-server-global-3.4/index.js
+++ b/packages/mongodb-memory-server-global-3.4/index.js
@@ -1,6 +1,9 @@
 'use strict';
 function __export(m) {
-  for (var p in m) if (!exports.hasOwnProperty(p)) exports[p] = m[p];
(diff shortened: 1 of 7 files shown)


nodkz:
@murpjack Thanks! 
May I ask you to make the same changes for `mongodb-memory-server-global-*` packages.

PS. It quite strange that `default` does not in its own properties.

murpjack:
Of course! I’ll make the changes this evening. I was having an error using typescript. Great product though 🏎

Jack

hasezoey:
squash merged, because [commit messages were not the required ones](https://github.com/nodkz/mongodb-memory-server/blob/master/.github/CONTRIBUTING.md#commit-structure) (angular-style/conventional-commits) and because these commits were changing the same thing across multiple files

PR #389, opened 2020-09-26
fix/improveMongoInstance

### LINKED ISSUE #166: "MongoMemoryReplSet.waitUntilRunning" returning before all ready
I'm trying to use the replicaset feature from mongodb-memory-server.

With this snippet:
```ts
const test = require('ava');
const { MongoMemoryReplSet } = require('mongodb-memory-server');
const { MongoClient } = require('mongodb');

const mongoReplSet = new MongoMemoryReplSet({
  debug: true,
  replSet: {
    count: 3,
    configSettings: { electionTimeoutMillis: 500 }
  }
});

test.before('Init', async () => {
  await mongoReplSet.waitUntilRunning();

  process.env.MONGODB_URI = `${await mongoReplSet.getConnectionString()}?replicaSet=testset`;
  console.log('process.env.MONGODB_URI:', process.env.MONGODB_URI);

  process.env.MONGODB_DBNAME = await mongoReplSet.getDbName();
  console.log('process.env.MONGODB_DBNAME:', process.env.MONGODB_DBNAME);

  const mongoClient = await MongoClient.connect(process.env.MONGODB_URI, { useNewUrlParser: true });
  const dbConn = mongoClient.db(process.env.MONGODB_DBNAME);

  dbConn.on('close', () => logger.warn('MongoDB event `close`'));
  dbConn.on('error', (err) => logger.warn('MongoDB event `error`:', err)); // eslint-disable-line promise/prefer-await-to-callbacks
  dbConn.on('fullsetup', () => logger.info('MongoDB event `fullsetup`'));
  dbConn.on('reconnect', () => logger.info('MongoDB event `reconnect`'));
  dbConn.on('timeout', () => logger.warn('MongoDB event `timeout`'));
});

test('Success!', (t) => {
  t.truthy(true);
});
```

I get:
```txt
  MongoError {
    [Symbol(mongoErrorContextSymbol)]: {},
    message: 'no primary found in replicaset or invalid replica set name',
  }
```

### LINKED ISSUE #365: MongoInstance: handle instance not being primary anymore
detect when the current instance is not the primary anymore

currently: only detects when the instance gets primary, but not when it is not anymore

(`MongoInstance.stdoutHandler`)

hasezoey:
- rename interface "MongodOps" to "MongodOpts" to keep naming consistency
- remove static value "childProcessList" because it was never used
- move "dynamic" function "debug" into private class function
- shorten "constructor" to use less lines
- outsource "MongodOpts.instance" into own interface "MongoInstanceOpts"
- change root values of "MongodOpts" to be required
- remove unused return value inside "run"'s "launch" Promise
- add tsdoc to "kill"
- add tsdoc to "errorHandler"
- extend tsdoc for "closeHandler"
- remove value "opts"
- add value "instanceOpts"
- add value "binaryOpts"
- add value "spawnOpts"
- prepareCommandArgs: remove splitting "this.instanceOpts"
- prepareCommandArgs: convert values to booleans (to ensure no falsy values)
- improve tsdoc for "waitPrimaryReady"
- change "MongoInstance" to extend "EventEmitter"
- add various Events
- track event names with an enum
- handle not being PRIMARY anymore
- "_waitForPrimary" use new events instead of function
- remove function "waitPrimaryReady"
- remove value "waitForPrimaryResolveFns"
- replace instanceReady with Event "instanceReady"
- replace instanceFailed with Event "instanceError"
- remove case for "shutting down with code"
- update case for "address already in use"
- add listener to "run" on "instanceClosed"
- test the error value on reject
- change test name to be more readable and accurate
- feat(db_util): add function "assertion"
- add error if "post" is null/undefined in "prepareCommandArgs" (because that is the purpose if this package)
- add error if "dbPath" is null/undefined in "prepareCommandArgs" (because that is the purpose if this package)
- move "ip" inside an if
- enforce "stdio" to be always "pipe"
- try "SIGKILL" if process didnt exit within 10 seconds
- reject promise if process dosnt exit after sending "SIGKILL"
- rename parameter "process" to "childProcess" to not overwrite global "process"
- move function "kill_internal" from "MongoInstance" to "db_util"
- rename function "kill_internal" to "killProcess"
- log the array created in the function
- overwrite "emit", "on", "once" to only allow "MongoInstanceEvents" instead of "string | symbol"
- add "@fires" to all functions that emit an event
- test that "kill" dosnt call "killProcess" if no childProcesses are running
- test that "_launchMongod" throws if "childProcess.pid" is undefined
- test that "errorHandler" emits "instanceRawError" and "instanceError"
- test that "stderrHandler" emits "instanceSTDERR"
- test that "stdoutHandler" emits correct event on "waiting for connections"
- move all "event testing" tests into "test events"
- add "beforeAll" creating the spy & instance & events map
- add "beforeEach" resetting spy-calls & events map
- test that "stdoutHandler" emits correct event on "address already in use"
- test that "stdoutHandler" emits correct event on "CURL_OPENSSL_3"
- test that "stdoutHandler" emits correct event on "CURL_OPENSSL_4"
- test that "stdoutHandler" emits correct event on "transition primary complete"
- test that "stdoutHandler" emits correct event on "state change"
- add "TODO" for events in "stdoutHandler" in "MongoInstance" that arnt covered yet
- remove redundant "debug.enabled" check
- rename event "instanceState" to "instanceReplState"
- remove "| null" from "childProcess"
- remove "| null" from "killerProcess"
- change "childProcess" to be optional
- change "killerProcess" to be optional
- reset "childProcess" reference to "undefined" in "stop"
- reset "killerProcess" reference to "undefined" in "stop"
- test(replset-multi): remove variable "opts"
- add tsdoc to the variables in "Mongoinstance"

## Related Issues

- closes #365
- most likely fixes #166 

---

because this pr includes breaking changes, i thought to bring this together with other file-improvements

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
@@ -67,8 +67,8 @@ export default class MongoInstance extends EventEmitter {
   binaryOpts: MongoBinaryOpts;
   spawnOpts: SpawnOptions;
 
-  childProcess: ChildProcess | null = null;
-  killerProcess: ChildProcess | null = null;
+  childProcess?: ChildProcess;
+  killerProcess?: ChildProcess;
   isInstancePrimary: boolean = false;
   isInstanceReady: boolean = false;
 
--- a/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
@@ -179,11 +179,13 @@ export default class MongoInstance extends EventEmitter {
 
     if (!isNullOrUndefined(this.childProcess)) {
       await killProcess(this.childProcess, 'childProcess');
+      this.childProcess = undefined; // reset reference to the childProcess for "mongod"
     } else {
       this.debug('- childProcess: nothing to shutdown, skipping.');
     }
     if (!isNullOrUndefined(this.killerProcess)) {
       await killProcess(this.killerProcess, 'killerProcess');
+      this.killerProcess = undefined; // reset reference to the childProcess for "mongo_killer"
     } else {
       this.debug('- killerProcess: nothing to shutdown, skipping.');
     }
--- a/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
@@ -20,7 +20,7 @@ if (lt(process.version, '10.15.0')) {
 
 const log = debug('MongoMS:MongoInstance');
 
-export interface MongodOps {
+export interface MongodOpts {
   // instance options
   instance: {
     port?: number;
@@ -44,7 +44,7 @@ export interface MongodOps {
  */
 export default class MongoInstance {
   static childProcessList: ChildProcess[] = [];
-  opts: MongodOps;
+  opts: MongodOpts;
   debug: DebugFn;
 
   childProcess: ChildProcess | null;
@@ -55,7 +55,7 @@ export default class MongoInstance {
   instanceReady: EmptyVoidCallback = () => {};
   instanceFailed: ErrorVoidCallback = () => {};
 
-  constructor(opts: MongodOps) {
+  constructor(opts: MongodOpts) {
     this.opts = opts;
     this.childProcess = null;
     this.killerProcess = null;
@@ -83,7 +83,7 @@ export default class MongoInstance {
    * Create an new instance an call method "run"
    * @param opts Options passed to the new instance
    */
-  static async run(opts: MongodOps): Promise<MongoInstance> {
+  static async run(opts: MongodOpts): Promise<MongoInstance> {
     const instance = new this(opts);
     return instance.run();
   }
--- a/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
@@ -5,7 +5,7 @@ import MongoBinary from './MongoBinary';
 import { MongoBinaryOpts } from './MongoBinary';
 import { StorageEngineT, SpawnOptions } from '../types';
 import debug from 'debug';
-import { isNullOrUndefined } from './db_util';
+import { assertion, isNullOrUndefined } from './db_util';
 import { lt } from 'semver';
 import { EventEmitter } from 'events';
 
@@ -67,9 +67,9 @@ export default class MongoInstance extends EventEmitter {
 
   constructor(opts: Partial<MongodOpts>) {
     super();
-    this.instanceOpts = opts.instance ?? {};
-    this.binaryOpts = opts.binary ?? {};
-    this.spawnOpts = opts.spawn ?? {};
+    this.instanceOpts = { ...opts.instance };
+    this.binaryOpts = { ...opts.binary };
+    this.spawnOpts = { ...opts.spawn };
 
     this.on(MongoInstanceEvents.instanceReady, () => {
       this.isInstanceReady = true;
@@ -109,17 +109,23 @@ export default class MongoInstance extends EventEmitter {
    * Create an array of arguments for the mongod instance
    */
   prepareCommandArgs(): string[] {
+    assertion(
+      !isNullOrUndefined(this.instanceOpts.port),
+      new Error('"instanceOpts.port" is required to be set!')
+    );
+    assertion(
+      !isNullOrUndefined(this.instanceOpts.dbPath),
+      new Error('"instanceOpts.dbPath" is required to be set!')
+    );
     const result: string[] = [];
-    result.push('--bind_ip', this.instanceOpts.ip || '127.0.0.1'); // default on all falsy values
     // "!!" converts the value to an boolean (double-invert) so that no "falsy" values are added
-    if (!!this.instanceOpts.port) {
-      result.push('--port', this.instanceOpts.port.toString());
-    }
+    result.push('--port', this.instanceOpts.port.toString());
+    result.push('--dbpath', this.instanceOpts.dbPath);
     if (!!this.instanceOpts.storageEngine) {
       result.push('--storageEngine', this.instanceOpts.storageEngine);
     }
-    if (!!this.instanceOpts.dbPath) {
-      result.push('--dbpath', this.instanceOpts.dbPath);
+    if (!!this.instanceOpts.ip) {
+      result.push('--bind_ip', this.instanceOpts.ip);
     }
     if (this.instanceOpts.auth) {
       result.push('--auth');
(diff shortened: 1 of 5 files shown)


hasezoey:
updated commit messages to use proper type (`feat`, `refactor`)

hasezoey:
force push to update commit messages to integrate `BREAKING CHANGE:`

packages/mongodb-memory-server-core/src/util/MongoInstance.ts
```diff
@@ -87,90 +104,97 @@ export default class MongoInstance {
...
      */
-    async function kill_internal(process: ChildProcess, name: string, debug: DebugFn) {
-      await new Promise((resolve) => {
-        const timeout = setTimeout(() => {
-          debug('kill_internal timeout triggered');
-          resolve();
-        }, 1000 * 10);
-        process.once(`exit`, (code, signal) => {
-          debug(`- ${name}: got exit signal, Code: ${code}, Signal: ${signal}`);
+    async function kill_internal(this: MongoInstance, childprocess: ChildProcess, name: string) {
```
nodkz, line 170: Better to move out this function to utils.
This helps to 
- remove `this` and `call` call
- add more readability (removing function in function)

packages/mongodb-memory-server-core/src/util/MongoInstance.ts
```diff
@@ -254,72 +266,86 @@ export default class MongoInstance {
...
+   * @param code The Exit code to handle
    */
-  closeHandler(code: number): void {
+  private closeHandler(code: number): void {
     if (code != 0) {
       this.debug('Mongod instance closed with an non-0 code!');
     }
     this.debug(`CLOSE: ${code}`);
-    this.instanceFailed(`Mongod instance closed with code "${code}"`);
+    this.emit(MongoInstanceEvents.instanceClosed, code);
```
nodkz, line 292: EvenEmitter my love! How I didn't add it before? Brilliant improvement!

packages/mongodb-memory-server-core/src/util/__tests__/MongoInstance-test.ts
```diff
@@ -66,8 +62,6 @@ describe('MongodbInstance', () => {
       },
     });
     expect(inst.prepareCommandArgs()).toEqual([
-      '--bind_ip',
-      '127.0.0.1',
```
nodkz, line 70: How I remember there was some problem in old mongod versions that didn't start without IP binding. Anyway, let's try to launch it without an explicit IP address. Of course better to start on all interfaces. If we are wrong then somebody opens an issue ;)

packages/mongodb-memory-server-core/src/util/db_util.ts
```diff
@@ -31,4 +31,16 @@ export function isNullOrUndefined(val: unknown): val is null | undefined {
   return val === null || val === undefined;
 }
 
+/**
+ * Assert an condition, if "false" throw error
+ * Note: it is not named "assert" to differentiate between node and jest types
+ * @param cond The Condition to throw
+ * @param error An Custom Error to throw
+ */
+export function assertion(cond: unknown, error?: Error): asserts cond {
```
nodkz, line 40: In most repos (especially from Facebook) I saw `invariant` name for the method of such kind
https://github.com/graphql/graphql-js/blob/master/src/jsutils/invariant.js

I don't know but maybe you saw `assertion` name in some other OSS repos which are used not in test suites.

nodkz (approved):
There are two small notes with requested changes. But we are ready to go without them ;)
So much really cool improvements. 👍 Especially EE ❤️

packages/mongodb-memory-server-core/src/util/__tests__/MongoInstance-test.ts
hasezoey: from my knowledge (what i read on the docs for mongod) is that mongod can start without any argument, and for bind_ip it will default to `localhost`

packages/mongodb-memory-server-core/src/util/db_util.ts
hasezoey: actually, i didnt read that somewhere - i created that function while reviewing the new `asserts` keyword in typescript (i think 3.7 or 3.8 added this), but because nodejs's assert didnt include this (and still dosnt), i had to create an new function and to not confuse (or accidentally import the wrong one) i named it `assertion`, instead of `assert` (and i didnt use nodejs's assert in that function, because i also tried to make typegoose "brower-ready" (running without nodejs))

PS: your linked file / function dosnt have the keyword `assertion`

nodkz: Yep, instead `assertion` name in the majority of repos is used `invariant` name. 
For example https://www.npmjs.com/package/invariant
Mostly it's used for code modification in PROD env in browsers. But we are on the server side and don't have aims to minify our code. So let keep this method with your name `assertion`.

packages/mongodb-memory-server-core/src/util/__tests__/MongoInstance-test.ts
nodkz: Yep, I also read the same. But in practice, I had some problems, but I don't remember what exactly the issue was. And maybe it's already gone or I missed something a couple of years ago. Anyway let's keep your solution and if some problems occur we in the first place will look at this place.

packages/mongodb-memory-server-core/src/util/MongoInstance.ts
hasezoey: should i also change the `this.debug` to have an parameter (and log as `MongoMS:MongoInstance`) or should it log as `MongoMS:db_utils`?

hasezoey: for now i made it to `MongoMS:db_utils` and renamed the function to `killProcess` (instead of `kill_internal`)

hasezoey:
force-push to rebase on latest beta (and removing the commits already included in 6.9.0)

hasezoey:
force-push to rebase on latest beta (fix coverage for beta, at least i hope)

hasezoey:
i think the pr is fully ready now

nodkz:
Monumental work! 👍
I reviewed it again and everything looks good!
You can ship when it necessary.

PS. I'm also almost finished with graphql-compose-mongoose [v9.0.0](https://github.com/graphql-compose/graphql-compose-mongoose/blob/alpha/docs/releases/9.0.0.md). I hope that can publish it today.

packages/mongodb-memory-server-core/src/util/MongoInstance.ts
nodkz: I think it will be better to use `MongoMS:utils`. But anyway I don't have strong opinion in this thing.

hasezoey: ok, it should always match the file name, i will rename it to `utils` after the PR(s)

PR #391, opened 2020-09-28
fix/improveMongoMemoryReplSet

hasezoey:
- add more tsdoc
- refactor values for "_state" into an enum
- use "isNullorUndefined" over "!= null"
- use "const" instead of "let" for "dbName"
- remove shortening named "rsOpts"
- remove "async" from "getDbName"
- refactor constructor to be more readable and shorter
- add tsdoc to "autoStart"
- add function "ensureAsync"
- "waitUntilRunning": throw error if state is not "running" or "init"
- "waitUntilRunning": ensure that the function is async
- add error to constructor if "replSet.count" is "<= 0" (and add test for it)
- if in state "init" "getUri" now uses "waitUntilRunning" because when "_waitForPrimary" returns it is not ensured that the state changed
- call "await ensureAsync" inside "start"
- change "if not state 'stopped'" to switch
- add more logs
- replace templating string with normal where not needed
- add link on why "ensureAsync" is needed
- test that "waitUntilRunning" throws an error if state is not "running" or "init"
- test that "getUri" actually throws if state is not "running" (or "init")
- test that "getUri" executes "waitUntilRunning" when state is "init"
- test that "start" actually throws if state is not "stopped"
- add test to test "instanceOpts"
- test that "waitUntilRunning" should return if state is "running"
- refactor multiple "if" into one switch
- test that "_initReplSet" throws an error if state is not "init"
-  _initReplSet: throw error if "this.servers.length" is "<= 0"
- test that "_initReplSet" throws if server count is 0 or less
- add log to "stop"
- remove function "getConnectionString"
- add function "stateChange"
- add overwrite for "on", "emit", "once"
- add function "create"
- remove option "autoStart"
- remove test to test option "autoStart"
- remove "?" from "MongoMemoryReplSetOptsT"
- change "_state" to be "protected"
- add getter "state"
- remove function "getDbName"
- rename "opts.instanceOpts" to "_instanceOpts"
- rename "opts.binary" to "_binaryOpts"
- rename "opts.replSet" to "_replSetOpts"
- add getter & setter "instanceOpts"
- add getter & setter "binaryOpts"
- add getter & setter "replSetOpts"
- remove option "oplogSize"
- change "getInstanceOpts" to be "protected"
- change "_initReplSet" to be "protected"
- change "_initServer" to be "protected"
- change "_waitForPrimary" to be "protected"
- remove value "_waitForPrimary" because it is only used once
- change "getUri" to be sync
- style: remove "uri" value when only used once (tests)
- add more tsdoc
- remove once used value "server"
- remove value "cnt"
- change "log" to give more information
- feat(MongoInstance): add value "isReplSet"
- _initReplSet: rename "conn" to "con"
- remove dynamic import "mongodb" (because on top of the file "import * mongodb" is already executed)
- feat(MongoInstance): graceful ReplSet shutdown (no "SIGKILL" usage anymore)
- change package "mongodb" to be non-optional
- dependencies(uuid): add "^"
- chore(commitlint.config): add "release" and "dependencies" as valid types
- _initReplSet: directly use db "admin"
- _initReplSet: remove redundant object assignment
- change if-error to "assertion"
- remove commented out HACK
- waitUntilRunning: shorten function
- test(replset-single): fix test to test "waitUntilRunning" waiting for event
---

[this documentation](https://snyk.io/blog/nodejs-how-even-quick-async-functions-can-block-the-event-loop-starve-io/) will help to understand why `setImmediate` is in some places


hasezoey:
force-push to fix broken commit

hasezoey:
force-push to clean up some commit's content

hasezoey:
force-push to move test to appropriate file

hasezoey:
this pr will remain WIP until #389 is merged

packages/mongodb-memory-server-core/src/MongoMemoryReplSet.ts
```diff
@@ -168,10 +225,17 @@ export default class MongoMemoryReplSet extends EventEmitter {
...
+      //   return this.waitUntilRunning();
+      case MongoMemoryReplSetStateEnum.stopped:
+        break;
+      case MongoMemoryReplSetStateEnum.running:
+      default:
+        throw new Error('Already in "init" or "running" state. Use debug for more info.');
     }
-    this.emit((this._state = 'init'));
+    this.emit((this._state = MongoMemoryReplSetStateEnum.init)); // this needs to be executed before "setImmediate"
+    await ensureAsync();
```
nodkz, line 238: I'll read about this tomorrow.
Thanks for the link https://snyk.io/blog/nodejs-how-even-quick-async-functions-can-block-the-event-loop-starve-io/ 🙏

nodkz (approved):
Very clear PR! 👍
Just one small thing – `MongoMemoryReplSetStateEnum` name from Java world 😅
But I've created a bit loooonger https://github.com/graphql-compose/graphql-compose/blob/master/src/ObjectTypeComposer.d.ts#L77 😜

hasezoey:
> Just one small thing – MongoMemoryReplSetStateEnum name from Java world sweat_smile

i just tried to keep the naming schema of the surrounding code (class name + what it is)

packages/mongodb-memory-server-core/src/MongoMemoryReplSet.ts
hasezoey: just to give an small example:
imagine running `start`, but not awaiting it and instead move on to `waitUntilRunning` and await that:
```ts
const replSet = new MongoMemoryReplSet();
replSet.start();
```
but before the promise of `start` (with `ensureAsync`) would get executed, the promise of `waitUntilRunning` (without `ensureAsync`) will be executed first, and so the error of being in the `stopped` state would come, before `start` even had the chance to change it
-> and `start` needs to have `ensureAsync` to execute surrounding code, if not being awaited (like for spinning up multiple replSets or setting up event listeners(after the call))
(Note: this might not fully represent what is happening, but should give an "clear" example of what this is trying to prevent)

hasezoey:

hasezoey:
i think the pr is ready now

main focus changed:
- no more "using SIGKILL" warnings (also test speed-up) / graceful replset shutdown
- better insurance against "infinite running / infinite waiting"

nodkz (approved):

hasezoey:

PR #394, opened 2020-09-29
Fix/improveMongoMemoryServer

hasezoey:
- remove double check
- remove "null" use "undefined"
- convert one if to optional-chaining
- remove unnecessary "else" path
- test that "stop" returns "true" if not running
- test that "getDbPath" returns the dbPath
- add "assertion"
- change "getPort" to be sync
- change "getDbPath" to be sync
- change "getDbName" to be sync
- (singleDB) remove call to "getConnectionString" and use ".create"
- (multipleDB) remove call to "getConnectionString" and use ".create"
- remove deprecated function "getConnectionString"
- change "getUri" to be sync
- change "getInstanceInfo" to return undefined
- remove call to "ensureInstance" inside "stop"
- remove option "autoStart"
- change that "create" now always calls ".start"
- remove test "should create an instance but not autostart"
- refactor "start" to be more readable
- merge "runningInstance" and "instanceInfoSync" into "instanceInfo"
- test that "ensureInstance" returns early if already running
- move test "getUri" with "otherDb" into MongoMemoryServer-test
- test that "getUri" returns generated otherDb
- (MongoMemoryServer-test) refactor file to be more readable
- add assertion for sanity to check if "instanceInfo.instance" is defined
- always reset "port" to "undefined" if instance failed and not just when "defined"
- killProcess: fix "SIGINT"-"SIGKILL" warn condition
- remove "await" from "getUriBase" call
- remove comment & change log
- remove "MongoInstanceDataT.childProcess"
- remove "StartupInstanceData.uri"
- shorten "getUri"
- rename function "assertionInstanceInfoSync" to "assertionInstanceInfo"
- change "instanceOpts" to be readonly and Readonly
- change "binaryOpts" to be readonly and Readonly
- change "spawnOpts" to be readonly and Readonly
- remove resetting "port" inside "start"
- test that "getUri" returns correct value if no parameter is provided
- start: remove first ".catch"
- test(MongoMemoryServer): change test names to represent the tests
- test(multipleDB): await connection.close
- test(multipleDB): move "beforeAll" and "afterAll" hooks inside describe
- test(multipleDB): move "db1" and "db2" inside only using test
- test(singleDB): move "beforeAll" and "afterAll" hooks inside describe
- test(singleDB): await connection.close & move "db" inside test
- stop: reset "servers" after stopping them
- remove "stop" call where "start" is mocked
- change "MongoMemoryServer" to extend "EventEmitter"
- add enum "MongoMemoryServerStateEnum"
- add enum "MongoMemoryServerEventEnum"
- add function "stateChange"
- ensureInstance: fix starting calling "start" even when "start" was already called
- test that "ensureInstance" throws if "instanceInfo" is "undefined" but "_state" is "running"
- test that "ensureInstance" throws if "_state" has no case
- test that "ensureInstance" throws an error if _state is "starting" but emitting not "running"
- test that "ensureInstance" calls "start" and starts normally
- change "instanceInfo" to be "protected"
- rename "instanceInfo" to "_instanceInfo"
- rename function "getInstanceInfo" to "instanceInfo"
- change "instanceInfo" to be an getter instead of an function
- change "_state" to be "protected"
- add getter function "state" to get the current state (return "_state")
- remove function "getPort" in favor of ".instanceInfo.port"
- remove function "getDbPath" in favor of ".instanceInfo.dbPath"
- remove function "getDbName" in favor of ".instanceInfo.dbName"

### DIFF
--- a/packages/mongodb-memory-server-core/src/__tests__/singleDB-test.ts
+++ b/packages/mongodb-memory-server-core/src/__tests__/singleDB-test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
 import { MongoClient } from 'mongodb';
 import MongoMemoryServer from '../MongoMemoryServer';
 
@@ -25,7 +26,7 @@ describe('Single mongoServer', () => {
   });
 
   it('should start mongo server', async () => {
-    const db = con.db(mongoServer.getDbName());
+    const db = con.db(mongoServer.instanceInfo!.dbName);
 
     expect(db).toBeDefined();
     const col = db.collection('test');
--- a/packages/mongodb-memory-server-core/src/__tests__/singleDB-test.ts
+++ b/packages/mongodb-memory-server-core/src/__tests__/singleDB-test.ts
@@ -2,31 +2,32 @@ import { Db, MongoClient } from 'mongodb';
 import MongoMemoryServer from '../MongoMemoryServer';
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 600000;
-let con: MongoClient;
-let db: Db;
-let mongoServer: MongoMemoryServer;
-
-beforeAll(async () => {
-  mongoServer = await MongoMemoryServer.create();
-  const mongoUri = mongoServer.getUri();
-  con = await MongoClient.connect(mongoUri, {
-    useNewUrlParser: true,
-    useUnifiedTopology: true,
-  });
-
-  db = con.db(mongoServer.getDbName());
-});
-
-afterAll(async () => {
-  if (con) {
-    con.close();
-  }
-  if (mongoServer) {
-    await mongoServer.stop();
-  }
-});
 
 describe('Single mongoServer', () => {
+  let con: MongoClient;
+  let db: Db;
+  let mongoServer: MongoMemoryServer;
+
+  beforeAll(async () => {
+    mongoServer = await MongoMemoryServer.create();
+    const mongoUri = mongoServer.getUri();
+    con = await MongoClient.connect(mongoUri, {
+      useNewUrlParser: true,
+      useUnifiedTopology: true,
+    });
+
+    db = con.db(mongoServer.getDbName());
+  });
+
+  afterAll(async () => {
+    if (con) {
+      con.close();
+    }
+    if (mongoServer) {
+      await mongoServer.stop();
+    }
+  });
+
   it('should start mongo server', async () => {
     expect(db).toBeDefined();
     const col = db.collection('test');
--- a/packages/mongodb-memory-server-core/src/__tests__/singleDB-test.ts
+++ b/packages/mongodb-memory-server-core/src/__tests__/singleDB-test.ts
@@ -7,7 +7,7 @@ let db: Db;
 let mongoServer: MongoMemoryServer;
 
 beforeAll(async () => {
-  mongoServer = new MongoMemoryServer();
+  mongoServer = await MongoMemoryServer.create({ autoStart: true });
   const mongoUri = await mongoServer.getUri();
   con = await MongoClient.connect(mongoUri, {
     useNewUrlParser: true,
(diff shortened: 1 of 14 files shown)


hasezoey:
force-push to rebase on latest beta & remove commits already included in 6.9.0 / previous PR #389

hasezoey:
`MongoMemoryServer` now extends `EventsEmitter` (like `MongoInstance`) to mainly solve that `ensureInstance` dosnt call `start` again if `start` was already called

hasezoey:
i think the pr is fully ready now

PS: the coverage somehow thinks the base is way before pr #389 got merged (even though base is 7.0.0-beta.1)

nodkz:
I'm not sure about making `getPort`, `getDbPath` synchronous. And if no port or path, we start throwing an Error. 

Before v7, if somebody asks a port, then under the hood lib begins to stat a mongod process. And the user receives the port of the launched mongod process.

Anyway, I think we should try your solution. And add some note to the readme that URI or port can be obtained only after starting a mongod process.

Amazing PR 👍

nodkz (approved):

hasezoey:
> Before v7, if somebody asks a port, then under the hood lib begins to stat a mongod process. And the user receives the port of the launched mongod process.

yes i understand, but i thought that the users of this package should have more control on when to start instead of having `ensureInstance` called in every function
-> later i would like to remove these functions, in favor of `getInstanceInfo` and maybe modify that function to be an getter and `instanceInfo` to be `protected` (not visible & modifiable outside the class (types only))

>  And add some note to the readme that URI or port can be obtained only after starting a mongod process.

sorry about that, im not at that part yet - i currently change "so much" and i would like to write the actual documentation and then remove the "currrent documentation" from the readme
-> TL;DR: i will change it when all the "important" changed are made

---

if we would remove these functions in favor of an getter, i could add an deprecation warning to 6.9 and the new way for an "easy transition", but i didnt (and currently dont) know if it would be good to change it that way

hasezoey:
the following commits changed what i proposed in my previous comment (`later i would like to remove these functions, in favor of getInstanceInfo and maybe modify that function to be an getter and instanceInfo to be protected (not visible & modifiable outside the class (types only))`)
- ec0ad2df2b4c7d44d48b7d76d4f0b20eb4e35f58
- f8f64a9e3dc2e3583b15d31631f3dd84350c2448
- 3d9c5e5f29c2b9e85a2aae63ef4c689f09fcf79d
- 42e74e511b1fbdb93e6323ea8ee2ad25a51bc291
- 2d89991ab5be3f92592cb6b5a523691975afa42b
- 95ba41081d44200e18724ba7ef8e514662e638e6

packages/mongodb-memory-server-core/src/__tests__/singleDB-test.ts
```diff
@@ -1,33 +1,33 @@
...
+      await con.close();
+    }
+    if (mongoServer) {
+      await mongoServer.stop();
+    }
+  });
 
-describe('Single mongoServer', () => {
   it('should start mongo server', async () => {
+    const db = con.db(mongoServer.instanceInfo!.dbName);
```
nodkz, line 29: The exclamation mark is evil 😈 Honestly, I don't remember why exactly I'm avoiding it in my code. But let it keep in the test suites.

nodkz (approved):
Getter `instanceInfo` looks good. 👍

packages/mongodb-memory-server-core/src/__tests__/singleDB-test.ts
hasezoey: i mean i know and understand why, but in the tests *you literally expect* the value to be defined (and not null / undefined)

nodkz: Yep, I understand this. My comment was just a note. Feel free to merge this PR.

hasezoey: yes i know, i will rebase this pr then merge it (when im at it again)

hasezoey:
force-push to rebase onto latest beta & reword some commits

hasezoey:
cc053c7f78ff29f67aabae30d180376c26ca96d8 reverts commit e0bc6764920fce8e08da4885af78528a3e77d433., because i noticed that `tsconfig.test` was not unused (it is used by eslint to not error on files outside of `packages/*core/src`)

PR #396, opened 2020-10-04
Fix/removeUnused

hasezoey:
- change "tsconfig.build" to extend "tsconfig"
- add script "build:tests"
- use tslib & "importHelpers"
- remove unused file "deprecate"
- fix(types): remove unused types
- fix(types): remove alias "SpawnOptions"
- change files to not only default export, but also export named-export
- refactor(index): directly export instead of import-export

---

"core" package size before tslib: `~368kb` after adding tslib: `~336kb` (size taken from before the commit and after the commit)

---

the coverage somehow thinks the base is way before pr #389 got merged (even though base is 7.0.0-beta.1)


hasezoey:

packages/mongodb-memory-server-core/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "rootDir": "./src",
-    "outDir": "./lib",
+    "outDir": "./build",
```
nodkz, line 4: Why do you change `outDir` in `tsconfig.json` to `./build`?

Firstly I had thought that this is an error, and the core package will be with empty `/lib` folder, but after that, I saw that you override it back to `./lib` in `tsconfig.build.json`.

packages/mongodb-memory-server-core/package.json
```diff
@@ -60,6 +60,7 @@
     "semver": "^7.3.2",
     "tar-stream": "^2.1.4",
     "tmp": "^0.2.1",
+    "tslib": "^2.0.1",
```
nodkz, line 63: We should try to keep `dependencies` as small as possible. I think that `importHelpers` with `tslib` are redundant even they help to reduce bundle size. 

`mongodb-memory-server` works on the backend, so bundle size is less critical than the number of prod dependencies.

packages/mongodb-memory-server-core/src/MongoMemoryServer.ts
```diff
@@ -46,7 +46,7 @@ export interface MongoInstanceDataT extends StartupInstanceData {
   childProcess?: ChildProcess;
 }
 
-export default class MongoMemoryServer {
+export class MongoMemoryServer {
```
nodkz, line 49: 👍
Named exports work very well with TS in VSCode. When you start typing the class name it can be auto-imported for you (will be added `import ...` in the file header).

packages/mongodb-memory-server-core/tsconfig.json
hasezoey: because this is the config, where everything `.ts` is included (like source AND Tests), but because we dont want the tests in the published build (and because we already had and tsconfig.build) i changed it in this one to `build` (so that it is not going into the same folder)

packages/mongodb-memory-server-core/package.json
hasezoey: from my knowledge, many projects use this when they also use typescript - so if i understand the chain correctly, tslib will only be installed once (if the versions match up)

-> this in general help keeping the node_modules folder smaller on disk (you have probably seen these jokes about it already)

nodkz (approved):

hasezoey:

Issue #379, opened 2020-09-16
SyntaxError: Invalid regular expression: /[\p{Ll}]/: Invalid escape


rvillane:
## Versions

- NodeJS: 10.12.0
- mongodb-memory-server-*: 6.7.6
- mongodb: 3.6.2
- mongoose: 5.10.5
- system: Linux Ubuntu

package: mongo-memory-server 

## What is the Problem?

getting this error when installing the latest version in Jenkins after upgrading from mongodb-memory-server@6.6.6 to mongodb-memory-server@6.7.6

```
> mongodb-memory-server@6.7.6 postinstall /home/jenkins/workspace/etail-transaction-service_master/node_modules/mongodb-memory-server
> node ./postinstall.js

/home/jenkins/workspace/etail-transaction-service_master/node_modules/mongodb-memory-server-core/node_modules/camelcase/index.js:17
} else if (isLastCharUpper && isLastLastCharUpper && /[\p{Ll}]/u.test(character)) {
^

SyntaxError: Invalid regular expression: /[\p{Ll}]/: Invalid escape
at preserveCamelCase (/home/jenkins/workspace/etail-transaction-service_master/node_modules/mongodb-memory-server-core/node_modules/camelcase/index.js:17:33)
at Object.camelCase [as default] (/home/jenkins/workspace/etail-transaction-service_master/node_modules/mongodb-memory-server-core/node_modules/camelcase/index.js:63:11)
at resolveConfig (/home/jenkins/workspace/etail-transaction-service_master/node_modules/mongodb-memory-server-core/lib/util/resolve-config.js:43:217)
at Object. (/home/jenkins/workspace/etail-transaction-service_master/node_modules/mongodb-memory-server-core/lib/util/resolve-config.js:56:15)
at Module._compile (module.js:653:30)
at Object.Module._extensions..js (module.js:664:10)
at Module.load (module.js:566:32)
at tryModuleLoad (module.js:506:12)
at Function.Module._load (module.js:498:3)
at Module.require (module.js:597:17)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! mongodb-memory-server@6.7.6 postinstall: `node ./postinstall.js`
npm ERR! Exit status 1

```

## Do you know *why* it happenes?

*no*

hasezoey:
[...]

PR #397, opened 2020-10-04
chore: change "target" to "es6"

hasezoey:
change compile target to `es6`

plus sides:
- reduces build size from `~360kb` to `~312kb` (core package)
- easier debug on problems (because the compiled code more resembles the source code)

down sides:
- not being able to run this package in an es5 engine anymore (this is so-or-so already thanks to `engines` and depended packages)

---

this is an separate PR from #396 because this might be an "major" change and debatable (at least from what i observed in typegoose)
main things i could observe (even with runtime warnings & documentation):
issues about "invalid syntax" errors (users tried to run typegoose in an non es6 environment (like nodejs before 10.15) (like #379 in this repository)

PS: documentation (like the other PR's for 7.0) will be added later

### DIFF
--- a/packages/mongodb-memory-server-core/tsconfig.json
+++ b/packages/mongodb-memory-server-core/tsconfig.json
@@ -3,7 +3,7 @@
     "rootDir": "./src",
     "outDir": "./build",
     "noEmit": false,
-    "target": "es5",
+    "target": "es6",
     "module": "commonjs",
     "sourceMap": true,
     "declaration": true,
--- a/packages/mongodb-memory-server-global-3.6/package.json
+++ b/packages/mongodb-memory-server-global-3.6/package.json
@@ -11,6 +11,9 @@
     "url": "https://github.com/nodkz/mongodb-memory-server.git",
     "directory": "packages/mongodb-memory-server-global-3.6"
   },
+  "engines": {
+    "node": ">=10.15.0"
+  },
   "homepage": "https://github.com/nodkz/mongodb-memory-server",
   "keywords": [
     "mongodb",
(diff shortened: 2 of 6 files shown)


hasezoey:

nodkz:
According to evergreen https://node.green/#ES2015-built-in-extensions-Array-prototype-methods-Array-prototype-values we can have problems only with Array.values with node below 10.8 but we dot use this method in our codebase.

Anyway, I think better to stay in es5. Reducing a bundle size is not our aim. Better to provide a seamless experience rather than introducing engine restrictions.

What do you think?

hasezoey:
> Anyway, I think better to stay in es5. Reducing a bundle size is not our aim. Better to provide a seamless experience rather than introducing engine restrictions.

true, but it also improves debugging (when having to look through the compiled code, at least) (and when using the nodejs debugger, less "renamed / extra values")

nodkz:
Do you often debug MMS at runtime?
If so let introduce this breaking change! V6 packages for old engines nobody will remove ;)

nodkz (approved):

nodkz:
PS. If you decide to merge this change, please check that `engine` proprty present package.json?

hasezoey:
> Do you often debug MMS at runtime?

might be an bad practice, but i prefere the "print-debug" strategy, but the debugger is really helpful when trying to get to know what keeps the process alive (like mongodb not getting killed)

---

i dont know if this comes from nodejs or vscode, but the debugger source-maps pretty well - but sometimes compiled output needs to be checked / is referred to (and anyway thanks to dependencies we use (like `camelCase`) nodejs before 10 will not work anymore, as seen in #379 )

hasezoey:
> PS. If you decide to merge this change, please check that engine property present in package.json.

at least defined in core's package.json at the moment, but will add the other packages as well

nodkz:
Let's migrate on es6!
Before I saw profit only in bundle size but now I'm thinking about our convenience.

hasezoey:

d41f0425b1a5ae2dbed2abf31c46364adc323474 added the `engines` field to all package.json

PR #404, opened 2020-10-11
getUriBase: remove trailing ?

caub:
I don't see the point for this trailing ?, there's no query strings added after, and it's making it harder to copy paste this url to for exmple connect to the DB (I have to `.replace(/\?$/, '')` in my code to make it easier, so I thought about a PR

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/db_util.ts
+++ b/packages/mongodb-memory-server-core/src/util/db_util.ts
@@ -20,7 +20,7 @@ export function getHost(uri: string): string {
  * Basic MongoDB Connection string
  */
 export function getUriBase(host: string, port: number, dbName: string): string {
-  return `mongodb://${host}:${port}/${dbName}?`;
+  return `mongodb://${host}:${port}/${dbName}`;
 }
 
 /**


hasezoey:
the `?` is there for functions using `getUriBase`, like `MongoMemoryServer.getUri()` and `MongoMemoryReplSet.getUri()`, and `getUriBase` is an *internal function* it is not meant to be used directly (use the functions described before)

caub:
Hi @hasezoey 
yes I know it's for .getUri(), but could we remove that '?' if there's nothing after? it's a bit easier to copy/paste from the console

For example, here's my current code
```
  const uri = await mongoServer.getUri();
  console.log('Using memory DB:', uri.replace(/\?$/, ''));
```

I think we could move that logic directly in getUri()?

hasezoey:
> yes I know it's for .getUri(), but could we remove that '?' if there's nothing after?

will look into that for 7.0 (this is at least an "minor" change (as in semver))

caub:
Ok thanks! and thanks for working on this great lib

(Let's close so)

PR #405, opened 2020-10-12
Fix/improvePostinstall

### LINKED ISSUE #174: mongodb-memory-server-global-3.6 downloads two mongos
seems like `mongodb-memory-server-global-3.6` does more work than neccessary.
package.json:
```
    "devDependencies": {
        "@types/mongodb-memory-server": "1.8.0",
        "mongodb-memory-server-global-3.6": "5.0.4",
    },

```

on my dev machine
```
$ npm install

> mongodb-memory-server@5.0.4 postinstall C:\project\package\node_modules\mongodb-memory-server
> node ./postinstall.js

mongodb-memory-server: checking MongoDB binaries cache...
mongodb-memory-server: binary path is C:\user\home\.cache\mongodb-binaries\4.0.3\mongod.exe

> mongodb-memory-server-global-3.6@5.0.4 postinstall C:\project\package\node_modules\mongodb-memory-server-global-3.6
> node ./postinstall.js

mongodb-memory-server: checking MongoDB binaries cache...
mongodb-memory-server: binary path is C:\user\home\.cache\mongodb-binaries\3.6.12\mongod.exe

> nodemon@1.18.7 postinstall C:\project\package\node_modules\nodemon
> node bin/postinstall || exit 0

```

in gitlab ci:
```
[INFO] --- frontend-maven-plugin:1.6:npm (npm-install) @ project-artifact-id ---
[INFO] Running 'npm install' in /builds/namespace/project/package
[INFO] 
[INFO] > mongodb-memory-server@5.0.4 postinstall /builds/namespace/project/package/node_modules/mongodb-memory-server
[INFO] > node ./postinstall.js
[INFO] 
[INFO] mongodb-memory-server: checking MongoDB binaries cache...
[INFO] Downloading MongoDB 4.0.3: 0 % (0mb / 79.4mb)
[INFO] mongodb-memory-server: binary path is /builds/namespace/project/package/node_modules/.cache/mongodb-memory-server/mongodb-binaries/4.0.3/mongod
[INFO] > mongodb-memory-server-global-3.6@5.0.4 postinstall /builds/namespace/project/package/node_modules/mongodb-memory-server-global-3.6
[INFO] > node ./postinstall.js
[INFO] 
[INFO] mongodb-memory-server: checking MongoDB binaries cache...
[INFO] Downloading MongoDB 3.6.12: 0 % (0mb / 114.2mb)
[INFO] mongodb-memory-server: binary path is /root/.cache/mongodb-binaries/3.6.12/mongod
```

### LINKED ISSUE #378: mongod is being installed twice with global
first with mongodb-memory-server and after with mongodb-memory-server-global

hasezoey:
- check if there is already an PRIMARY before running "_waitForPrimary"
- change all packages to depend on "-core"
- add "postinstall" file into "core"
- change errors to be better
- change "releaserc" to match new dependencies
- use "tslib.__exportStar" over custom "__export"
- change "mongodb_version" to latest patch version
- delete "mongodb-memory-server-global-3.4"
- delete "mongodb-memory-server-global-4.0"
- change "LATEST_VERSION" to latest patch version "4.0.20"
- move "postinstall" into "util"
- rename "postinstall" to "postinstall-helper"
- remove dependency "os" from all packages
- add dependency "os" to "-core"
- all postinstall.js: de-duplicate code

## Related Issues

- fixes #378 
- fixes #174


hasezoey:
coverage will decrease because no more `SIGKILL` warning gets executed (thanks to upgrading to 4.0.20)

packages/mongodb-memory-server-global-3.6/postinstall.js
```diff
@@ -1,19 +1,21 @@
 /* eslint @typescript-eslint/no-var-requires: 0 */
 
-function isModuleExists(name) {
+function doesModuleExist(name) {
```
nodkz, line 3: An interesting solution to move postinstall script to the core package 👍

In such case we can drastically simplify all `postinstall` scripts in`-global-` packages to the following form:
```js
var mongodb_version = require('./package.json').mongodb_version;
require('mongodb-memory-server-core/lib/utils/postinstall-helper.js').downloadBinaries({ version: mongodb_version });
```

There is no need 
- to require `setDefaultValue` method
- global directory can be determined in core package

`postinstall.js` better to rename to `./utils/postinstall-helper.js` for removing the misconception that `core` package has postinstall script and somebody forgets to call it )

nodkz (changes requested):

packages/mongodb-memory-server-global-3.6/postinstall.js
nodkz: So we can remove all copy-pasta from postinstall scripts and put there just two lines
- which version needs to install
- call the autoinstall method from core package

nodkz:
> coverage will decrease because no more SIGKILL warning gets executed (thanks to upgrading to 4.0.20)

Don't worry about coverage reports, they added for decorations ;) Like a reminder for contributors that will be nice to add some tests. But you are making incredible changes & improvements. And wasting your time for coverage – it's absolutely unnecessary thing.

.releaserc.js
```diff
@@ -45,11 +45,6 @@ module.exports = {
       "@google/semantic-release-replace-plugin",
       {
         "replacements": [
-          {
-            "files": ["packages/*/package.json"],
-            "from": "\"mongodb-memory-server\": \".*\"",
-            "to": "\"mongodb-memory-server\": \"${nextRelease.version}\"",
-          },
```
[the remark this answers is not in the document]
hasezoey: because, from my knowledge this was to replace `dependencies: { mongodb-memory-server: "x.x.x" }` with the new version, and because that package is not depended anymore (now all depend on `-core` to fix the mentioned issues) so i removed that replace (unnecessary)

hasezoey:
> Don't worry about coverage reports, they added for decorations ;) Like a reminder for contributors that will be nice to add some tests. But you are making incredible changes & improvements. And wasting your time for coverage – it's absolutely unnecessary thing.

just wanted to mention an "TL;DR" on why the coverage decreased (and not stayed the same or increased)

---

will apply the mention change (rename & move to "util") + the simplification

.releaserc.js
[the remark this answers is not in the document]
nodkz: Ah, ok.

So if all `global` packages now depends from `mongodb-memory-server-core` we need to replace that code on:
```diff
             "files": ["packages/*/package.json"],
-            "from": "\"mongodb-memory-server\": \".*\"",
+            "from": "\"mongodb-memory-server-core\": \".*\"",
-            "to": "\"mongodb-memory-server\": \"${nextRelease.version}\"",
+            "to": "\"mongodb-memory-server-core\": \"${nextRelease.version}\"",
```
for keeping in sync all global packages with new published version of core package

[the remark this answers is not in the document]
nodkz: Otherwise, all global packages would be frozen on incorrect `mongodb-memory-server-core` in their package.json files

packages/mongodb-memory-server-global-3.6/postinstall.js
hasezoey: 2a9aab7738e6512ce0e03135f4c8a8d217d62816 should have this applied

.releaserc.js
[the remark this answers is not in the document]
hasezoey: no, because that already existed before, so i only removed the one replacing `mongodb-memory-server:` (because that dosnt apply anymore)

look here for an reference: https://github.com/nodkz/mongodb-memory-server/blob/beta/.releaserc.js#L43-L60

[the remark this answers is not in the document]
nodkz: Sorry, I missed that!

packages/mongodb-memory-server-core/src/util/postinstall-helper.ts
```diff
@@ -23,11 +30,22 @@ if (typeof envSystemBinary === 'string') {
   process.exit(0);
 }
 
-(async () => {
+export async function postInstall(version?: string, local?: boolean): Promise<never | void> {
```
nodkz, line 33: Maybe `downloadBinaryToGlobal`? I dislike this name but it even better than `postInstall`.

hasezoey: `downloadBinaryToGlobal` wouldn't really apply, because it is not always downloading to global (parameter `local`)

packages/mongodb-memory-server/postinstall.js
```diff
@@ -18,11 +18,12 @@ function doesModuleExist(name) {
   }
 }
 
-const modulePath = '../mongodb-memory-server-core/lib/postinstall';
+const modulePath = '../mongodb-memory-server-core/lib/util/postinstall-helper';
```
nodkz, line 21: `mongodb-memory-server-core` in `dependncies` so I think that nodejs will correctly resolve `require('mongodb-memory-server-core/lib/util/postinstall-helper')` without relative import with doesModuleExist-check.

hasezoey: true

hasezoey: 6dc6343f03cd2f811825049a11454fe43cfb9aee and 6dc3222055ba099cad7f872f9ac814c45720b6b3 should apply this

packages/mongodb-memory-server-core/src/util/postinstall-helper.ts
nodkz: Maybe `postInstallDownload`, `postInstallEnsureBinary`?
This method should have some information about what it does. Just `postInstall` name is too general.

Anyway, you may keep your current method name if you dislike any names above and don't have any alternatives

packages/mongodb-memory-server/postinstall.js
hasezoey: i will revert 6dc6343f03cd2f811825049a11454fe43cfb9aee again, because from my testing (and test results) it is needed, because yarn will symlink the inter-workspace dependencies (and when doing an local install for testing both npm and yarn symlink)

hasezoey: i will need to revert the other one too, because tests (install) are failing

nodkz: Agreed. 

Honestly, I'm shocked that it didn't resolve correctly

Screen Shot 2020-10-13 at 01 10 14

nodkz (approved):

packages/mongodb-memory-server-core/src/util/postinstall-helper.ts
hasezoey: using `postInstallEnsureBinary` in aca826211351e73458c56e6d0dce99d068e05941

packages/mongodb-memory-server/postinstall.js
hasezoey: i thought too that it would resolve correctly, but it seems like it dosnt

-> but i think in the tests the problem is, that yarn tries to run "postinstall" which depends on that the code it compiled, which it isnt at that point, we could fix this with [`--ignore-scripts`](https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-ignore-scripts), downside would be other packages scripts wouldnt be executed too

nodkz: I think that we should stay on the current workaround via importing by the relative path with doesModuleExist-check.

PR #407, opened 2020-10-14
Fix/various

### LINKED ISSUE #299: How to specify authentication username, password and database when starting with `auth` set to `true`
## Versions

- NodeJS: 12.13.1
- mongodb-memory-server-*: 6.5.2
- mongodb: 4.2.2
- mongoose: 5.9.7
- system: Win10

package: mongo-memory-server 

## What is your question?

I am using this package in my tests. My actual code has my mongoose connecting to a secured db with credentials. So, I am getting an authentication failed error when I am starting the mongo memory server without auth and mongoose in my code passes in a username and password along with the test database.

Now, I know I can work around this by changing how the module which handles mongoose connection behave differently during test(not pass the username and password) but I saw that there is an option to start mongo memory server with `auth`. So, when I enable `auth` and try to connect with mongoose **without** auth, it gets connected but operations are not authzd, as expected.

But if I want mongoose to connect to the mongo memory server instance which has auth enabled, how do I specify the username, password and db?

Or should I drop this approach.

### LINKED ISSUE #374: Refactor "MongoBinary.cache" to Map<string, string>
is there maybe an reason on why it is not an Map / Record already?

### LINKED ISSUE #395: Unify Interface Names
currently there are interfaces named with `T` and without `T`, should we unify them and set an structure for it?

CC @nodkz

### LINKED ISSUE #406: Remove file "types" and put all types in the appropriate files
Move types like:
- `DownloadProgress` into `MongBinaryDownload`
- `MongoMemoryInstancePropBase` into `MongoInstance`
- `MongoMemoryInstanceProp` into `MongoInstance`
- `StorageEngine` into ?

hasezoey:
- replace "npm" with "yarn" to run scripts
- rename function "getUriBase" to "uriTemplate"
- only add "?" when query options are provided
- add test file "db_util-test"
- remove test "v6.0.0 adds "?" to the connection string (uri)"
- add regex to test the message of the error to ensure that is the only error ignored
- rename all test files from "-test" to ".test"
- change "cache" to be an Map
- change "getCachePath" return type to "string | undefined"
- remove default for variable "binaryPath"
- change interfaces & types to have an unified naming-style (remove "T")
- add section "Code Styles" to CONTRIBUTING
- uriTemplate: change "query" to be an string-array
- remove file "types.ts"
- move all types from "type.ts" into appropriate files
- rename interface "DownloadProgress" to "MongoBinaryDownloadProgress"
- style(MongoInstance): add TODO comments
- rename to match other file names [camelCase]
  - rename file "db_util" to "utils"
  - rename file "db_util.test" to "utils.test"
  - rename variables "db_util" to "utils"
  - rename file "resolve-config" to "resolveConfig"
  - rename file "resolve-config.test" to "resolveConfig.test"
  - rename file "postinstall-helper" to "postinstallHelper"
- chore(CONTRIBUTING): add section "File Names"
- test(MongoBinaryDownloadUrl): fix spies & mocks
- feat(MongoMemoryServer): add ability to automatically create auth
- feat(MongoMemoryReplSet): add basic "createAuth" support
- fix(utils): killProcess: check if the childProcess PID is still alive
- feat(MongoMemoryServer): add (protected) function "getNewPort"
- ensureInstance: move "return instanceInfo" into case "running"
- feat(MongoMemoryServer): add function "getStartOptions"
- refactor(MongoMemoryServer): start: check "state" instead of "instanceInfo"
- move test from "singleDB" to "MongoMemoryServer"
- feat(MongoMemoryServer): allow re-use of instances & dbPath
- fix(MongoInstance): run: reset all booleans on
- feat(MongoMemoryServer): add ability to force same port on restart
- fix(MongoMemoryReplSet): _waitForPrimary: check if instance is already primary
- feat(MongoMemoryReplSet): getUri: allow executing while state is "init"
- initAllServers: execute "stateChange" with "init"
- feat(MongoMemoryReplSet): allow re-use of instances & dbPath

## Related Issues

- closes #404 
- closes #374 
- closes #395 
- closes #406 
- closes #299


nodkz (approved):

nodkz:
💅 good polishing
👍

packages/mongodb-memory-server-core/src/util/db_util.ts
```diff
@@ -25,8 +25,8 @@ export function getHost(uri: string): string {
 /**
  * Basic MongoDB Connection string
  */
-export function getUriBase(host: string, port: number, dbName: string): string {
-  return `mongodb://${host}:${port}/${dbName}?`;
+export function uriTemplate(host: string, port: number, dbName: string, query?: string): string {
+  return `mongodb://${host}:${port}/${dbName}` + (!isNullOrUndefined(query) ? `?${query}` : '');
```
hasezoey, line 29: should i change `query` to be an array of strings, to do something like `uriTemplate(ip, port, dbName, ['option1=1', 'option=2'])` or should it stay `uriTemplate(ip, port, dbName, 'option1=1&option=2')`?

hasezoey:
Note: not merging yet, because im planning to include some more changes like:
- #406
- #365 
- #206 (and some other rsConfig options)
- maybe #299

packages/mongodb-memory-server-core/src/util/db_util.ts
nodkz: There is no purpose to do it right now. It can be done after publishing v7.0.0.

But if you want to do something crazy with a connection string, then you may see the following gist with parser/serializer https://gist.github.com/nodkz/d2bb4e91ba8ca9fd3178d45b33d94357

According to [_parseOptions mthod](https://gist.github.com/nodkz/d2bb4e91ba8ca9fd3178d45b33d94357#file-connectionstring-ts-L149) options are represented by Object, not array.

For example, we are using this parser for creating ENV variables for some systems which are not support connection strings, e.g for [redis-sentinel](https://gist.github.com/nodkz/753ef827d0083c377a944e0f7e468bcf)

hasezoey: this is an bit overkill for this project (at least at the moment), so i thought either `option1=1&option2=2` for now, and ask if it should be change to an array `['option1=1', 'option2=2']` (or maybe an object)
-> so if i understand correctly, i can let it stay like it is atm, and change when needed otherwise

nodkz: According to our current [config options](https://github.com/nodkz/mongodb-memory-server#available-options-for-mongomemoryserver) we accept `instance.args` as array. In such a way, `object` isn't good shape for us if we don't want to change its type.

So if we decide to provide static typings for all available instance's args then `object` will be good. BUT we work with different major mongod versions, which have different args. And making args statically typed can be an unsolved job. Now I come to a resolution that `object` for `options` is not suitable for us.

So `string` or `array of strings` is the best form for options. And how I see now an array of strings is more convenient.

hasezoey: ok, will change

> And making args statically typed can be an unsolved job

i think we only "officially support" the most common arguments (that are found on many versions, example: `auth`)

hasezoey: 8dffa64ee4d56523cefd993b1158d2a9e224cd8a changes this to an string-array

packages/mongodb-memory-server-core/src/__tests__/replset-single.test.ts
```diff
@@ -203,6 +203,83 @@ describe('single server replset', () => {
...
+    expect(users.users[0].user).toEqual(replSet.replSetOpts.auth.customRootName);
+    // @ts-expect-error
+    expect(MongoMemoryReplSet.prototype.initAllServers).toHaveBeenCalledTimes(1);
+    expect(console.warn).toHaveBeenCalledTimes(1);
+
+    await con.close();
+    await replSet.stop();
+  });
+
+  // TODO: re-enable this test when an solution for keeping dbPath across replset restarts
```
hasezoey, line 246: currently for ReplSet's the `createAuth` for non `ephemeralForTest` instances is not working, because dbPath (the temporary paths created by `MongoMemoryServer`) is not being kept across restarts of the same ReplSet instance

hasezoey:
i think the pr is ready now

nodkz:
Wow! Tremendous! 👍

nodkz:
I've just tried to test beta.5 with the new code and got the following error:
Screen Shot 2020-10-27 at 18 57 42

If we look at https://unpkg.com/browse/mongodb-memory-server-core@7.0.0-beta.5/ then we can see that folder `lib` is absent. I think we somewhere broke the build process. The problem was introduced in beta.2

nodkz:
You can merge this PR. 
But before publishing v7 we need to fix the build problem (in this or new PR).

hasezoey:
@nodkz thanks for noticing this, i think i know where the problem is

hasezoey:
@nodkz fixed it within 1e2f2fc51c81db0c58193416780f9b8bf578eecb

PR #412, opened 2020-10-27
Fix: Cannot find module 'mongodb-memory-server' in beta.6

nodkz:
Catch very strange behaviour with `export { *** } from ***;`. In transpiled code `MongoMemoryServer` does not defined anywhere.

https://unpkg.com/browse/mongodb-memory-server-core@7.0.0-beta.6/lib/index.js

Screen Shot 2020-10-27 at 21 35 48

This PR fix this problem.


hasezoey (approved):
it seems like `export {obj} from ""` was allowed to co-exit with `export default obj` (as in types) but actually not compiling as if working - thanks for catching this

hasezoey:
>  In transpiled code MongoMemoryReplSet does not defined anywhere.

are you sure you mean `MongoMemoryReplSet` not `MongoMemoryServer`?

nodkz:
I wanted to write `MongoMemoryServer`. Need to fix my T9 😜

PS. I've updated the issue description.

PR #415, opened 2020-11-02
Fix/improveCode

hasezoey:
- add enum for all resolveConfig Variables
- remove alias "reInitializePackageJson"
- remove value "LATEST_VERSION"
- set default version inside resolveConfig
- test(MongoBinaryDownload): replace "process.env.variable" with "process.env[values.variable]"
- refactor "should use direct download"
- fix multiline test name
- refactor "should skip download if binary tar exists"
- refactor "should pick up proxy from env vars"
- refactor "should not reject unauthorized"
- refactor "should reject unauthorized"
- refactor "makeMD5check returns true"
- refactor "makeMD5check throws an error"
- refactor "disables makeMD5check validation"
- makeMD5check: add more tsdoc
- test(MongoBinary): replace "process.env.variable" with "process.env[values.variable]"
- test(MongoBinary): refactor tests
- test(MongoInstance): refactor to use "isAlive"
- test(resolveConfig): use "fs.promises" instead of "promisify"
- test(replset-single): add description on why to use "ts-expect-error"
- test(replset-multi): directly use db "admin"
- test(MongoMemoryServer): add description on why to use "ts-expect-error"
- refactor(MongoBinary): replace "promisify" with "fs.promises"
- refactor(MongoBinaryDownload): replace "promisify" with "fs.promises"
- refactor(getos): replace "promisify" with "fs.promises"
- add function "pathExists"
- remove function "locationExists"
- test that "statPath" throws an error if code is not "ENOENT"
- test that assertion throws default error if none provided
- assertion: remove error code from default error
- refactor(MongoBinaryDownload): replace "existsSync" with "utils.pathExists"
- add more tsdoc to "statPath" & "pathExists"
- remove function "getCachePath" (because it just an alias to `cache.get`, which is public anyway)
- getSystemPath: also check for execute permission
- feat(MongoInstance): run: add check that the mongoBinary has sufficient permissions
- feat(MongoBinaryDownload): startDownload: add check that the downloadDir has sufficient permissions
- eslintrc
  - add environment "node"
  - set rule "no-unused-expressions" to warn
  - comma-dangle: change "functions" to "never"
  - set rule "no-else-return" to "warn"
  - add rule "padding-line-between-statements" for "return"
  - add rule "padding-line-between-statements" for "if"
  - add rule "padding-line-between-statements" for "function" & "class"
  - add rule "padding-line-between-statements" for "import"

---

[here](https://unix.stackexchange.com/a/317446/325185) is an good explanation on which permissions are needed for an directory for this package

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,7 +43,13 @@ module.exports = {
     '@typescript-eslint/explicit-function-return-type': 0,
     '@typescript-eslint/no-use-before-define': 0,
     '@typescript-eslint/no-empty-function': 0,
-    'curly': ['error', 'all']
+    'curly': ['error', 'all'],
+    'padding-line-between-statements': [
+      "warn",
+      {
+        blankLine: 'always', prev: '*', next: 'return'
+      }
+    ]
   },
   env: {
     node: true,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,7 +19,7 @@ module.exports = {
   rules: {
     'no-underscore-dangle': 0,
     'arrow-body-style': 0,
-    'no-unused-expressions': 0,
+    'no-unused-expressions': 1,
     'no-plusplus': 0,
     'no-console': 0,
     'func-names': 0,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -63,7 +63,16 @@ module.exports = {
       },
       {
         blankLine: 'always', prev: ['function', 'class'], next: '*' // add blank line *after* all functions and classes
-      }
+      },
+      {
+        blankLine: 'always', prev: '*', next: 'import' // add blank line *before* all imports
+      },
+      {
+        blankLine: 'always', prev: 'import', next: '*' // add blank line *after* all imports
+      },
+      {
+        blankLine: 'never', prev: 'import', next: 'import' // dont allow blank line between imports
+      },
     ]
   },
   env: {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -35,7 +35,7 @@ module.exports = {
     ],
     'no-prototype-builtins': 0,
     'prefer-destructuring': 0,
-    'no-else-return': 0,
+    'no-else-return': 1,
     'lines-between-class-members': ['error', 'always', { exceptAfterSingleLine: true }],
     '@typescript-eslint/explicit-member-accessibility': 0,
     '@typescript-eslint/no-explicit-any': 0,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -46,6 +46,7 @@ module.exports = {
     'curly': ['error', 'all']
   },
   env: {
+    node: true,
     jest: true,
   },
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -30,7 +30,7 @@ module.exports = {
         objects: 'always-multiline',
         imports: 'always-multiline',
         exports: 'always-multiline',
-        functions: 'ignore',
+        functions: 'never',
       },
     ],
     'no-prototype-builtins': 0,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -47,7 +47,16 @@ module.exports = {
     'padding-line-between-statements': [
       "warn",
       {
-        blankLine: 'always', prev: '*', next: 'return'
+        blankLine: 'always', prev: '*', next: 'return' // add blank line *before* all returns (if there are statements before)
+      },
+      {
+        blankLine: 'always', prev: '*', next: 'if' // add blank line *before* all ifs
+      },
+      {
+        blankLine: 'always', prev: 'if', next: '*' // add blank line *after* all ifs
+      },
+      {
+        blankLine: 'any', prev: 'if', next: 'if' // allow blank line between ifs, but not enforce either
       }
     ]
   },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -57,6 +57,12 @@ module.exports = {
       },
       {
         blankLine: 'any', prev: 'if', next: 'if' // allow blank line between ifs, but not enforce either
+      },
+      {
+        blankLine: 'always', prev: '*', next: ['function', 'class'] // add blank line *before* all functions and classes
+      },
+      {
+        blankLine: 'always', prev: ['function', 'class'], next: '*' // add blank line *after* all functions and classes
       }
     ]
   },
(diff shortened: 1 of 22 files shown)


README.md
```diff
@@ -286,6 +286,7 @@ MONGOMS_USE_LINUX_LSB_RELEASE # Only try "lsb_release -a"
 MONGOMS_USE_LINUX_OS_RELEASE # Only try to read "/etc/os-release"
 MONGOMS_USE_LINUX_ANYFILE_RELEASE # Only try to read the first file found "/etc/*-release"
 MONGOMS_ARCHIVE_NAME="mongodb-linux-x86_64-4.0.0.tgz" # Specify what file / archive to download
+SKIP_OS_RELEASE # ignore error thrown in "tryOSRelease"
```
nodkz, line 289: Suppose that it's a typo. I think it should be `MONGOMS_SKIP_OS_RELEASE`

hasezoey: yes definitely, this was an result of the copy-pasting i had done for the enum after i noticed this one was missing in the readme

hasezoey: fixed in e35249523a275fe7b83a31420aa3c0210349cb60

.eslintrc.js
```diff
@@ -30,22 +30,53 @@ module.exports = {
...
     'lines-between-class-members': ['error', 'always', { exceptAfterSingleLine: true }],
     '@typescript-eslint/explicit-member-accessibility': 0,
     '@typescript-eslint/no-explicit-any': 0,
     '@typescript-eslint/no-inferrable-types': 0,
     '@typescript-eslint/explicit-function-return-type': 0,
     '@typescript-eslint/no-use-before-define': 0,
     '@typescript-eslint/no-empty-function': 0,
-    'curly': ['error', 'all']
+    'curly': ['error', 'all'],
+    'padding-line-between-statements': [
```
nodkz, line 47: Wow-wow it's amazing 👍
Definitely will add the same rules in our internal projects! 😍

packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
```diff
@@ -75,15 +76,15 @@ export class MongoBinaryDownload {
...
 
-    if (await this.locationExists(mongodPath)) {
+    if (await pathExists(mongodPath)) {
       return mongodPath;
     }
 
     const mongoDBArchive = await this.startDownload();
     await this.extract(mongoDBArchive);
-    fs.unlinkSync(mongoDBArchive);
+    await promises.unlink(mongoDBArchive);
```
nodkz, line 85: Just a note: `promises` variable looks a little bit strange for me in raw code.
In our apps we are using the following style (if Node>v14): 
```
import fs from 'fs/promises';

await fs.unlink...
```
Or we are using such style for Node>v10 – `fs.promises.unlink` which is more clear for newcomers

But feel free to leave the current style (without `fs`) without changes.

packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
```diff
@@ -38,13 +38,16 @@ export class MongoBinaryDownloadUrl {
     const archive = await this.getArchiveName();
     log(`Using "${archive}" as the Archive String`);
 
-    const downloadUrl = resolveConfig('DOWNLOAD_URL');
+    const downloadUrl = resolveConfig(ResolveConfigVariables.DOWNLOAD_URL);
```
nodkz, line 41: For me looks overweighted. The old variant was better.

PS. Feel free to leave your new variant as is. It's not a big problem.

packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownload.test.ts
```diff
@@ -70,70 +77,72 @@ MONGOMS_MD5_CHECK environment variable`, () => {
...
   });
 
   it('should reject unauthorized when npm strict-ssl config is true', async () => {
     // npm sets true config value as string 'true' in env vars
     process.env['npm_config_strict_ssl'] = 'true';
 
     const du = new MongoBinaryDownload({});
-    du.httpDownload = jest.fn();
-    du.locationExists = jest.fn().mockReturnValue(false);
+    jest.spyOn(du, 'httpDownload').mockResolvedValue('/tmp/someFile.tgz');
```
nodkz, line 96: 👍

nodkz (approved):

nodkz:
As always, it's amazing work! 👍

Esp big tnx for eslint rule `lines-between-statements` 🙏

packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
hasezoey: i just made it so that there cannot be any spelling mistakes when using in multiple places, or having to rename / change the value to something to have it automatically in all places of this repo & for the user to easily change it

packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
hasezoey: i could also rename it, like i have done in `utils.ts` to `fspromises`

or just re-export these from `utils.ts`

.eslintrc.js
hasezoey: i searched for an rule similar to `newline-before-return` from tslint (to have consistent returns here) and the first one i found was deprecated into this rule which was more capable - so i also added more stuff an just the return thing

packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
nodkz: For this reason, we can use literal union type:
```ts
resolveConfig(name: 'DOWNLOAD_URL' | '...' | '...' | ...)
```
TS will control us for spelling mistakes. And renaming with TS also won't be a big deal.

packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
nodkz: fspromises also looks good

packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
hasezoey: > we can use literal union type

yes, but they wont be available at runtime (not that i would currently know on why it would be useful)

packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
hasezoey: fixed in f876670aaecc3b74ca4a85a80d804c137a40d572

PR #416, opened 2020-11-08
Fix/spawn

### LINKED ISSUE #193: [Windows] Can't stop mongod process
I'm running tests with mocha and I have the following code to stop the mongo server:

```ts
after(function (done) {
  this.timeout(10000);

  db.close()
    .then(() => mongoServer.stop())
    .then(() => { done(); })
    .catch((error) => { console.log('Error stopping mongo server', error); done(); });
  });
});
```

When I run the tests, I get an error message logged in the catch statement. I enabled debug mode and saw this output:

```
Called MongoMemoryServer.stop() method
Called MongoMemoryServer.ensureInstance() method:
Shutdown MongoDB server on port 55555 with pid 11724
Mongo[55555]: Called MongoInstance.kill():
Mongo[55555]:  - childProcess: send kill cmd...
Mongo[55555]:  - childProcess: got exit signal. Ok!
Mongo[55555]:  - killerProcess: send kill cmd...
Mongo[55555]: 2019-06-05T13:54:07.979-0700 I NETWORK  [conn1] end connection 127.0.0.1:55557 (3 connections now open)
2019-06-05T13:54:07.979-0700 I NETWORK  [conn2] end connection 127.0.0.1:55559 (2 connections now open)
2019-06-05T13:54:07.979-0700 I NETWORK  [conn3] end connection 127.0.0.1:55561 (1 connection now open)
2019-06-05T13:54:07.979-0700 I NETWORK  [conn4] end connection 127.0.0.1:55563 (0 connections now open)

Mongo[55555]: [MongoKiller]: exit - [null,"SIGTERM"]
Mongo[55555]:  - killerProcess: got exit signal. Ok!
Removing tmpDir C:\Users\UserName\AppData\Local\Temp\mongo-mem-16836ZeiScx7S058a
Error stopping mongo server { Error: EBUSY: resource busy or locked, unlink 'C:\Users\UserName\AppData\Local\Temp\mongo-mem-16836ZeiScx7S058a\mongod.lock'
    at Object.unlinkSync (fs.js:951:3)
    at _rmdirRecursiveSync (c:\path\to\project\node_modules\tmp\lib\tmp.js:309:12)
    at Object._cleanupCallback [as removeCallback] (c:\path\to\project\node_modules\tmp\lib\tmp.js:446:7)
    at MongoMemoryServer.<anonymous> (c:\path\to\project\node_modules\mongodb-memory-server-core\src\MongoMemoryServer.ts:162:14)
    at step (c:\path\to\project\node_modules\mongodb-memory-server-core\lib\MongoMemoryServer.js:43:23)
    at Object.next (c:\path\to\project\node_modules\mongodb-memory-server-core\lib\MongoMemoryServer.js:24:53)
    at fulfilled (c:\path\to\project\node_modules\mongodb-memory-server-core\lib\MongoMemoryServer.js:15:58)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

Am I doing something wrong here?

hasezoey:
- remove dependency `cross-spawn`, replace with `child_process`

this is because `cross-spawn` uses an extra shell to execute the binary in, which in result does not exit correctly (at least on windows)
replace `execSync` with `spawnSync` because of the same reason, `exec` & `execSync` use an extra shell too

-> the only drawback i can see is that the parent process dosnt have stdio with the mongo_killer process anymore (cannot be done without the parent waiting / killing the fork on exit)

## Related Issues

- fixes #193

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoInstance.ts
@@ -1,6 +1,5 @@
-import { ChildProcess, SpawnOptions } from 'child_process';
-import { default as spawnChild } from 'cross-spawn';
-import path from 'path';
+import { ChildProcess, fork, spawn, SpawnOptions } from 'child_process';
+import path, { resolve } from 'path';
 import MongoBinary from './MongoBinary';
 import { MongoBinaryOpts } from './MongoBinary';
 import debug from 'debug';
@@ -310,7 +309,7 @@ export class MongoInstance extends EventEmitter {
    * @fires MongoInstance#instanceLaunched
    */
   _launchMongod(mongoBin: string): ChildProcess {
-    const childProcess = spawnChild(mongoBin, this.prepareCommandArgs(), {
+    const childProcess = spawn(resolve(mongoBin), this.prepareCommandArgs(), {
       ...this.spawnOpts,
       stdio: 'pipe', // ensure that stdio is always an pipe, regardless of user input
     });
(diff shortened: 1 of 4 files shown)


packages/mongodb-memory-server-core/src/util/MongoInstance.ts
```diff
@@ -337,29 +336,16 @@ export class MongoInstance extends EventEmitter {
...
-        parentPid.toString(),
-        childPid.toString(),
-      ],
-      { stdio: 'pipe' }
+    const killer = fork(
+      path.resolve(__dirname, '../../scripts/mongo_killer.js'),
+      [parentPid.toString(), childPid.toString()],
+      {
+        detached: true,
+        stdio: 'ignore', // stdio cannot be done with an detached process cross-systems and without killing the fork on parent termination
```
nodkz, line 344: This is sad news.

Do we know what platforms have such problems? Maybe we if it just a win32 then provide killer without stdio support, but for other platforms keep full log/debug functionality.

Or you sure that everything will work without problems and checking stdout/err is redundant?

nodkz (approved):
Anyway, I approve of these changes. I think better to have a working solution than good logging in an unworking tool.

packages/mongodb-memory-server-core/src/util/MongoInstance.ts
hasezoey: > Do we know what platforms have such problems?

from what i know, its an nodejs quirk, not platform specific

> Or you sure that everything will work without problems and checking stdout/err is redundant?

im actually unsure if even before this change the stdio worked

nodkz: I can suggest a dirty solution for now: just comment the old code (do not remove completely) and keep the new one. This will make it easier to roll back the code in the far future.

hasezoey: couldnt we then just revert this commit?

hasezoey: fixed in 132917fac63745b351dfd73eb30f7f54962ddf37

we can just revert this commit if we ever need to

PR #424, opened 2020-12-16
feat(os-support): fall back to ubuntu release for arch-alikes

### LINKED ISSUE #302: [Arch] Distribution Support
When i set binary.version to '4.2.5' i get an error that it doesn't exist, not sure if that is an upstream issue or an issue with this library but 4.2.5 is definitely available.

### LINKED ISSUE #419: What MONGOMS_DOWNLOAD_MIRROR to use for Manjaro Linux
## Versions

- NodeJS: 10.16.1
- mongodb-memory-server-*: 6.6.7
- mongodb: 4.4.1
- mongoose: 5.10.10
- system: Manjaro Linux 20.2

```
$ node --experimental-repl-await
> const getos = require('getos')
> const { promisify } = require('util')
> let os = await promisify(getos)()
> console.log(os)
{ os: 'linux',
  dist: 'Manjaro Linux',
  codename: 'nibia',
  release: '20.2' }
```

package: mongo-memory-server-core

## What is your question?

I am on Manjaro Linux and I want to use mongo-memory-server. What `MONGOMS_DOWNLOAD_MIRROR` to use for Manjaro Linux?

boredland:
## Related Issues
- fixes #419
- fixes #302

I figured that the approach used in the aur mongodb-bin package (https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=mongodb-bin) should easily work on arch-alikes without much changes. For me it does.

### DIFF
--- a/packages/mongodb-memory-server-core/package.json
+++ b/packages/mongodb-memory-server-core/package.json
@@ -54,6 +54,7 @@
     "get-port": "^5.1.1",
     "https-proxy-agent": "^5.0.0",
     "lockfile": "^1.0.4",
+    "lookpath": "^1.1.0",
     "md5-file": "^5.0.0",
     "mkdirp": "^1.0.4",
     "mongodb": "^3.6.2",
--- a/packages/mongodb-memory-server-core/src/util/MongoBinary.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoBinary.ts
@@ -93,7 +93,7 @@ export class MongoBinary {
             ? `MongoBinary: Error when removing download lock ${err}`
             : `MongoBinary: Download lock removed`
         );
-        res(); // we don't care if it was successful or not
+        res(undefined); // we don't care if it was successful or not
       });
     });
 
--- a/packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
@@ -21,14 +21,10 @@ export class MongoBinaryDownloadUrl {
   platform: string;
   arch: string;
   version: string;
-  major: number;
-  minor: number;
   os: AnyOS | undefined;
 
   constructor({ platform, arch, version, os }: MongoBinaryDownloadUrlOpts) {
     this.version = version;
-    this.major = parseInt(this.version.split('.')[0], 10)
-    this.minor = parseInt(this.version.split('.')[1], 10)
     this.platform = this.translatePlatform(platform);
     this.arch = this.translateArch(arch, this.platform);
     this.os = os;
@@ -77,7 +73,7 @@ export class MongoBinaryDownloadUrl {
       case 'linux':
         return this.getArchiveNameLinux();
       default:
-        throw new Error(`Unkown Platform "${this.platform}"`);
+        throw new Error(`Unknown Platform "${this.platform}"`);
     }
   }
 
@@ -172,6 +168,7 @@ export class MongoBinaryDownloadUrl {
       console.warn(
         `There is no official build of MongoDB for ArchLinux (${os.dist}). Falling back to Ubuntu release.`
       );
+
       // falling back to ubuntu similar to https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=mongodb-bin
       return this.getUbuntuVersionString({
         os: 'linux',
@@ -232,6 +229,7 @@ export class MongoBinaryDownloadUrl {
     } else if (fedoraVer < 12 && fedoraVer >= 6) {
       name += '55';
     }
+
     return name;
   }
 
@@ -334,30 +332,26 @@ export class MongoBinaryDownloadUrl {
   getUbuntuVersionString(os: LinuxOS): string {
     let name = 'ubuntu';
     const ubuntuVer: string[] = os.release ? os.release.split('.') : [];
-    const majorVer: number = parseInt(ubuntuVer[0], 10);
-    // for all cases where its just "10.10" -> "1010"
-    // and because the "04" version always exists for ubuntu, use that as default
-    let version = `${majorVer || 14}04`;
+    const ubuntuMajor: number = parseInt(ubuntuVer[0], 10);
 
     if (os.release === '14.10') {
-      version = '1410-clang';
-    } else if (majorVer >= 18) {
-      if (this.major === 3) {
-        // For MongoDB 3.x using 1604 binaries, download distro does not have builds for Ubuntu 1804 AND 2004
+      return (name += '1410-clang');
+    } else if (ubuntuMajor >= 18) {
+      if (semver.satisfies(this.version, '3.x.x')) {
+        // there are no MongoDB 3.x binary distributions for ubuntu > 16.04
         // https://www.mongodb.org/dl/linux/x86_64-ubuntu1604
-        version = '1604';
-      } else if ((this.major === 4 && this.minor <= 3) || majorVer === 18) {
-        // See full list of versions https://www.mongodb.org/dl/linux/x86_64-ubuntu1804
-        // For MongoDB <4.4 using 1804 binaries, because 2004 doesn't have anything below 4.4
-        version = '1804';
-      } else if (majorVer >= 20) {
-        name += '2004';
+        return (name += '1604');
+      }
+      if (semver.satisfies(this.version, '<=4.3.x')) {
+        // there are no MongoDB 4.(x < 4) binary distributions for ubuntu > 18.04
+        // https://www.mongodb.org/dl/linux/x86_64-ubuntu1804
+        return (name += '1804');
       }
     }
 
-    name += version;
-
-    return name;
+    // for all cases where its just "10.10" -> "1010"
+    // and because the "04" version always exists for ubuntu, use that as default
+    return (name += `${ubuntuMajor || 14}04`);
   }
 
   /**
@@ -384,7 +378,7 @@ export class MongoBinaryDownloadUrl {
       case 'sunos':
         return 'sunos5';
       default:
-        throw new Error(`Unkown Platform "${platform}"`);
+        throw new Error(`Unknown Platform "${platform}"`);
     }
   }
 
--- a/packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
@@ -152,43 +152,46 @@ export class MongoBinaryDownloadUrl {
   getLinuxOSVersionString(os: LinuxOS): string {
     if (/ubuntu/i.test(os.dist)) {
       return this.getUbuntuVersionString(os);
-    } else if (/elementary OS/i.test(os.dist)) {
+    }
+    if (/elementary OS/i.test(os.dist)) {
       return this.getElementaryOSVersionString(os);
-    } else if (/suse/i.test(os.dist)) {
+    }
+    if (/suse/i.test(os.dist)) {
       return this.getSuseVersionString(os);
-    } else if (/rhel/i.test(os.dist) || /centos/i.test(os.dist) || /scientific/i.test(os.dist)) {
+    }
+    if (/rhel/i.test(os.dist) || /centos/i.test(os.dist) || /scientific/i.test(os.dist)) {
       return this.getRhelVersionString(os);
-    } else if (/fedora/i.test(os.dist)) {
+    }
+    if (/fedora/i.test(os.dist)) {
       return this.getFedoraVersionString(os);
-    } else if (/^linux\s?mint\s*$/i.test(os.dist)) {
+    }
+    if (/^linux\s?mint\s*$/i.test(os.dist)) {
       return this.getMintVersionString(os);
-    } else if (/debian/i.test(os.id_like || os.dist)) {
+    }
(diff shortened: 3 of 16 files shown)


packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
```diff
@@ -316,9 +325,11 @@ export default class MongoBinaryDownloadUrl {
         // For MongoDB 3.x using 1604 binaries, download distro does not have builds for Ubuntu 1804
         // https://www.mongodb.org/dl/linux/x86_64-ubuntu1604
         name += '1604';
-      } else {
-        // See fulllist of versions https://www.mongodb.org/dl/linux/x86_64-ubuntu1804
+      } else if (majorVer === 18) {
+        // See full list of versions https://www.mongodb.org/dl/linux/x86_64-ubuntu1804
```
boredland, line 329: was there a particular reason it stopped at ubuntu 18?

hasezoey (changes requested):
please re-target and rebase onto the beta branch, this should also answer your question

PS: thanks for the spelling fixes

hasezoey:
could you rebase again?

packages/mongodb-memory-server-core/src/util/getos/index.ts
```diff
@@ -125,9 +126,13 @@ async function getLinuxInfomation(): Promise<LinuxOS> {
  */
 async function tryLSBRelease(): Promise<LinuxOS | undefined> {
   try {
-    const lsb = await promisify(exec)('lsb_release -a'); // exec this for safety, because "/etc/lsb-release" could be changed to another file
+    // use upstream lsb file if it exists (like in linux mint)
```
boredland, line 129: imo this will prevent us from custom handling for some derivates (tried it for mint, according to google it should work for elementary too). therefore I extended the custom parsers for elementary and mint can be dropped.

boredland: Will try this in a vm later/tomorrow

packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts
```diff
@@ -350,69 +401,7 @@ describe('MongoBinaryDownloadUrl', () => {
       ).toBe('debian92');
     });
   });
-
-  describe('getMintVersionString', () => {
```
hasezoey, line 354: why did these tests get removed? it still could test if those distros are giving the correct thing - even though no dedicated function exists

packages/mongodb-memory-server-core/src/util/getos/index.ts
```diff
@@ -125,9 +126,13 @@ async function getLinuxInfomation(): Promise<LinuxOS> {
...
+    if (fs.existsSync('/etc/upstream-release/lsb-release')) {
```
hasezoey, line 130: if already testing for this file (that i have never seen before), maybe add an check for `/etc/lsb-release` before the command

packages/mongodb-memory-server-core/src/util/getos/index.ts
```diff
@@ -169,7 +174,7 @@ async function tryOSRelease(): Promise<LinuxOS | undefined> {
  */
 async function tryFirstReleaseFile(): Promise<LinuxOS | undefined> {
   try {
-    const file = (await fspromises.readdir('/etc')).filter(
+    const file = fs.readdirSync('/etc').filter(
```
hasezoey, line 177: why change this to be sync again?

packages/mongodb-memory-server-core/src/util/getos/index.ts
```diff
@@ -181,7 +186,7 @@ async function tryFirstReleaseFile(): Promise<LinuxOS | undefined> {
       throw new Error('No release file found!');
     }
 
-    const os = await fspromises.readFile(join('/etc/', file));
+    const os = fs.readFileSync(join('/etc/', file));
```
hasezoey, line 189: why change this to be sync again?

hasezoey (changes requested):
some suggestions

and thanks for all the spelling fixes

PS: im on LinuxMint, i could test it (but guessing from the code, it will work)

packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts
boredland: i removed them, because `MongoBinaryDownloadUrl` should never again get initialized with sth like 
```
 os: 'linux',	
 dist: 'Linux Mint',	
 release: '16',	
 ```
since it will take the values from linux mints upstream lsb which contains ubuntu properties.

File: packages/mongodb-memory-server-core/src/util/getos/index.ts (hunk 4 above)
Comment (boredland) [reply packages/mongodb-memory-server-core/src/util/getos/index.ts:130]: according to this [comment](https://github.com/nodkz/mongodb-memory-server/pull/424/files#diff-730c93549b867e9788e59c98aa3caf807c4b4d07012c9d2e2dc04b05ee32bf1eR134), that you already worked around by using lsb_release instead of the file itself...

File: packages/mongodb-memory-server-core/src/util/getos/index.ts (hunk 5 above)
Comment (boredland) [reply packages/mongodb-memory-server-core/src/util/getos/index.ts:177]: good question. no real reason, perhaps a bit consistency. are there real differences between using the sync methods and a promise that you immediately await for?

File: packages/mongodb-memory-server-core/src/util/getos/index.ts (hunk 4 above)
Comment (hasezoey) [reply packages/mongodb-memory-server-core/src/util/getos/index.ts:130]: sorry, what comment? for me this link goes to somewhere random in the changes in the pr (without an line marking or any comment near other than codecov)

File: packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts (hunk 3 above)
Comment (hasezoey) [reply packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts:354]: and what is when an other method is forced?
https://github.com/nodkz/mongodb-memory-server/tree/beta#options-which-can-be-set-via-environment-variables
> ```sh
> MONGOMS_USE_LINUX_LSB_RELEASE
> MONGOMS_USE_LINUX_OS_RELEASE
> MONGOMS_USE_LINUX_ANYFILE_RELEASE
> ```

File: packages/mongodb-memory-server-core/src/util/getos/index.ts (hunk 5 above)
Comment (hasezoey) [reply packages/mongodb-memory-server-core/src/util/getos/index.ts:177]: in this case not really, but in 7.0.0 i tried to make things less "sync-blocking"

Comment (boredland) [reply packages/mongodb-memory-server-core/src/util/getos/index.ts:177]: I read up a little and don't see any apparent downsides in using either of those (other than the taste-argument) in this place. I'd go with whatever you prefer. I you want me to revert, I'll do that.

Comment (boredland) [reply packages/mongodb-memory-server-core/src/util/getos/index.ts:177]: And I'm a bit confused. Wasn't that using the `promisify` util before? If so, perhaps I wanted to remove that not realizing that you already did 🎉

Comment (hasezoey) [reply packages/mongodb-memory-server-core/src/util/getos/index.ts:177]: like i said, in this case it dosnt really matter, but i tried to use less sync things wherever possible

Comment (hasezoey) [reply packages/mongodb-memory-server-core/src/util/getos/index.ts:177]: > And I'm a bit confused. Wasn't that using the `promisify` util before? If so, perhaps I wanted to remove that not realizing that you already did tada

yes, in 6.x it was using this, but since 7.0 (betas) i moved it to use https://nodejs.org/api/fs.html#fs_fs_promises_api

Comment (boredland) [reply packages/mongodb-memory-server-core/src/util/getos/index.ts:177]: will revert.

File: packages/mongodb-memory-server-core/src/util/getos/index.ts (hunk 4 above)
Comment (boredland) [reply packages/mongodb-memory-server-core/src/util/getos/index.ts]: hm. in the next line simply:
```
    // exec this for safety, because "/etc/lsb-release" could be changed to another file
```

File: packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts (hunk 3 above)
Comment (boredland) [reply packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts:354]: sooo. I think i made them a bit nicer, but after all I readded the custom handlers for elementary and mint.

File: packages/mongodb-memory-server-core/src/util/getos/index.ts [hunk 7]
```diff
@@ -127,8 +127,10 @@ async function getLinuxInformation(): Promise<LinuxOS> {
 async function tryLSBRelease(): Promise<LinuxOS | undefined> {
   try {
     // use upstream lsb file if it exists (like in linux mint)
-    if (await fspromises.stat('/etc/upstream-release/lsb-release')) {
-      return parseLSB(await fspromises.readFile('/etc/upstream-release/lsb-release', 'utf8'));
+    if (await pathExists('/etc/upstream-release/lsb-release')) {
```
boredland, line 130: didn't know that existed.

packages/mongodb-memory-server-core/src/util/getos/index.ts
```diff
@@ -209,9 +211,9 @@ async function tryFirstReleaseFile(): Promise<LinuxOS | undefined> {
 export function parseLSB(input: string): LinuxOS {
   return {
     os: 'linux',
-    dist: input.match(LSBRegex.name)?.[1] ?? 'unknown',
-    codename: input.match(LSBRegex.codename)?.[1],
-    release: input.match(LSBRegex.release)?.[1] ?? '',
+    dist: input.match(LSBRegex.name)?.[2] ?? 'unknown',
```
boredland, line 214: regex - killing me since 1999

packages/mongodb-memory-server-core/package.json
```diff
@@ -54,6 +54,7 @@
     "get-port": "^5.1.1",
     "https-proxy-agent": "^5.0.0",
     "lockfile": "^1.0.4",
+    "lookpath": "^1.1.0",
```
boredland, line 57: added this to check if a command is present. I think its handy, but of course there are more methods to check that.

packages/mongodb-memory-server-core/src/util/getos/index.ts
```diff
@@ -71,151 +74,118 @@ async function getLinuxInfomation(): Promise<LinuxOS> {
...
  * Try the "lsb_release" command, and if it works, parse it
  */
-async function tryLSBRelease(): Promise<LinuxOS | undefined> {
-  try {
-    const lsb = await promisify(exec)('lsb_release -a'); // exec this for safety, because "/etc/lsb-release" could be changed to another file
-
-    return parseLSB(lsb.stdout);
-  } catch (err) {
-    // check if "USE_LINUX_LSB_RELEASE" is unset, when yes - just return to start the next try
-    if (isNullOrUndefined(resolveConfig(ResolveConfigVariables.USE_LINUX_LSB_RELEASE))) {
```
boredland, line 133: as I see it, it is safe to return a function/promise here instead of catching stuff etc.

the way this currently works is:
if an env var forces one of the methods, that method gets wrapped in a try/catch and then some more spaghetti checks if forcing was enabled and if so, throws the original error. otherwise returns void.

the way this now would work:
if an env var forces one of the methods, that methods gets return and it may throw if it fails.
in the only place where we want to suppress errors, when collecting the info from different sources and take one that worked, we can suppress it as we please.

hasezoey: i think of removing these variables, i once thought they could be useful, but from my knowledge it just adds code clutter and didnt provide anything useful - will remove it in some 7.0 beta

hasezoey:
@boredland is this pr ready to be reviewed again / merged?

boredland:
> @boredland is this pr ready to be reviewed again / merged?

I'd think so.

hasezoey:
ran tests on my machine (Linux Mint 20), first the in `getos` (`should return ubuntu info`) failed, reason being that it somehow gets `Linuxmint` instead of `Ubuntu`

my guess is that `getLSBRelease` gets executed, but somehow `jest.spyOn(fsPromises, 'stat').mockRejectedValueOnce({ code: 'ENOENT' })` has no effect, and `/etc/upstream-release/lsb-release` gets used, and *somehow* the mock implementation ignores the error on the last line (from `jest.spyOn(fsPromises, 'readFile')`), and reads the real file

packages/mongodb-memory-server-core/src/util/MongoBinary.ts
```diff
@@ -93,7 +93,7 @@ export class MongoBinary {
             ? `MongoBinary: Error when removing download lock ${err}`
             : `MongoBinary: Download lock removed`
         );
-        res(); // we don't care if it was successful or not
+        res(undefined); // we don't care if it was successful or not
```
hasezoey, line 96: why add an `undefined` here? the result of this promise is not used

packages/mongodb-memory-server-core/src/util/errors.ts
```diff
@@ -0,0 +1,4 @@
+export class FileNotFoundError {
```
hasezoey, line 1:
```suggestion
export class FileNotFoundError extends Error {
```

shouldnt it extend an normal `Error`?

hasezoey:
closing in favor of #432

PR #426, opened 2021-01-05
chore: upgrade mongodb to 3.6.3

ostrowr:
Bump mongodb dependency, and loosen version constraint from exact to `^`

When installing mongodb-memory-server alongside another library that installs mongodb (e.g. mongoose) using a tool like npm that doesn't have as fine-grained control over dependencies as does yarn, you end up with several instances of mongodb and it's a nightmare to get mongoose to work. This PR just bumps the package version to 3.6.3 (latest) and allows compatible versions to be installed via `^` semantics.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "jest": "26.4.2",
     "lerna": "^3.22.1",
     "lint-staged": "10.4.0",
-    "mongodb": "3.6.2",
+    "mongodb": "^3.6.3",
     "prettier": "2.1.2",
     "semantic-release": "^17.1.2",
     "ts-jest": "26.4.1",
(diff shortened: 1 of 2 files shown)


hasezoey:
> Bump mongodb dependency, and loosen version constraint from exact to ^

these are only `dev-dependencies`, they do not affect outside code (or runtime code at all), they dont prevent an higher version being used, look at `peerDepedencies` instead

ostrowr:
> these are only dev-dependencies, they do not affect outside code (or runtime code at all), they dont prevent an higher version being used, look at peerDepedencies instead

This isn't generally true, depending on the npm setup. In any case, is there a reason to keep mongodb at an old version?

hasezoey:
> This isn't generally true,

could you elaborate on this one?

> In any case, is there a reason to keep mongodb at an old version?

no, but this is just installed for tests, and as an note, this is *the workspace* package file, which is not included in any of the packages - and aside from that, did you already look at the beta version of this repository? because there, it dosnt exist with in "workspace level" anymore, but at the package level (as an normal dependency (thanks to some features) [and also with an `^`]


TL;DR: i also closed the PR, because this was already changed in the beta

PR #454, opened 2021-03-03
feat: change binary from "base/version/mongod" to "base/mongod-arch-dist-version"

### LINKED ISSUE #256: Add distro and version to the mongod file
(this enhancement is taken from a conversation at #254 with @firstnoises)

## Why?

when a user (like @firstnoises pointed out), uses the local system and lets say a docker system with a different distro, to be able to use the normal package (non global), and not having to re-download all the time

hasezoey:
## Related Issues

- fixes #256


hasezoey:
@nodkz is this ok to be changed, or should it stay like it was?

nodkz:
More details is good.
Moreover if somebody uses heterogeneous systems it helps avoid unproper binary using.

PR #459, opened 2021-03-16
fix(MongoMemoryServer): change state to stopped when start fails

idan-at:
When the server fails to start, the state is left is `running`. 

changing the state to `stopped` allows implementing more complex scenarios over the testkit. 
For example, in case the testkit failed to find a proper binary, I can set the `MONGOMS_DOWNLOAD_URL` environment variable and try again.

ATM, doing so throws because `start` [expects](https://github.com/nodkz/mongodb-memory-server/blob/beta/packages/mongodb-memory-server-core/src/MongoMemoryServer.ts#L256) the state to either be `new` or `stopped`

### DIFF
--- a/packages/mongodb-memory-server-core/src/MongoMemoryServer.ts
+++ b/packages/mongodb-memory-server-core/src/MongoMemoryServer.ts
@@ -277,6 +277,8 @@ export class MongoMemoryServer extends EventEmitter {
         console.warn('Starting the instance failed, enable debug for more information');
       }
 
+      this.stateChange(MongoMemoryServerStates.stopped);
+
       throw err;
     });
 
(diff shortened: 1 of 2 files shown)


idan-at:
@hasezoey can you please have a look?

Although coverage has decreased, we are covered by the test I updated.

hasezoey:
> Although coverage has decreased

if you look closely, you would notice that the coverage report is not right, i already made an issue with codecov about this ~4 months ago, but still didnt get it fixed
-> in this project, coverage is not necessary

hasezoey (approved):
thanks for this fix, i didnt notice this

idan-at:
Thanks for merging!

Do you minding releasing a new beta RC?

hasezoey:
> Do you minding releasing a new beta RC?

its automatic after the tests pass (semantic-release)

Issue #440, opened 2021-02-19
Config paths from `package.json` are relative to the working directory instead of `package.json`


mhassan1:
## Versions

- NodeJS: 14.15.1
- mongodb-memory-server-*: 6.9.3
- mongodb: 4.4.1
- mongoose: 5.11.15
- system: MacOS

package: `mongodb-memory-server-core`

## What is the Problem?

When `findPackageJson` grabs configuration from a `package.json`, it sets `downloadDir` and `systemBinary` to the relative paths it finds in `package.json`. When `MongoBinary` is looking for the binary, it will look in those locations relative to the current working directory; however, it's possible that the current working directory will vary depending where the initial command was run. For example, I may run a test from the root of my project, or I may run it from the test directory. If I have configuration in `package.json` in the root, the binary will be resolved differently in those scenarios.

## Code Example

We could add code to `findPackageJson` that resolves `downloadDir` and `systemBinary` to absolute paths (based on the location of the `package.json`):
```ts
const ourConfig = value?.config?.mongodbMemoryServer || {};

// resolve relative paths
for (const relativePathProp of ['downloadDir', 'systemBinary']) {
  if (ourConfig[relativePathProp]) {
    ourConfig[relativePathProp] = path.resolve(path.dirname(filename), ourConfig[relativePathProp]);
  }
}
packageJsonConfig = ourConfig;
```

## Do you know *why* it happenes?

Once a relative `downloadDir` is provided to `mongoBinaryDownload`, it is resolved relative to the current working directory, for example: https://github.com/nodkz/mongodb-memory-server/blob/2bc51280ec91ac8558c2aa49d3995d0d499f3620/packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts#L99

hasezoey:
commit 0bec0bf1b63904e82e357dc6dc646db3c0fa6df4 should resolve this, could you confirm? (version `7.0.0-beta.17`)

mhassan1:
LGTM, thank you!

mhassan1:
I've been doing more testing of this, and I'm noticing that the fix in https://github.com/nodkz/mongodb-memory-server/commit/0bec0bf1b63904e82e357dc6dc646db3c0fa6df4 is incomplete. The `packageJsonConfig[ccDownloadDir]` value passed to `path.resolve()` should be resolved relative to the found `package.json` file. Instead, it is relative to the current working directory.

[...]

PR #469, opened 2021-05-06
fix(resolveConfig): resolve download dir and system binary relative to found package.json

mhassan1:
Resolves https://github.com/nodkz/mongodb-memory-server/issues/440


hasezoey:
thanks, i didn't notice that i didn't do this

PR #502, opened 2021-07-06
fix: improve error message `"instanceInfo" is undefined` for users who upgraded from v6

nodkz:
I've just migrated `graphql-compose-mongoose@7.1.0` to a new version and met with the following problems:
- got unfriendly `"instanceInfo" is undefined` error message which does not explain what happens and what to do. In this PR I tried to provide users a little bit more information on how to resolve this problem.
- I found that README.md contains old examples for v6. I've changed docs for the first example. But I didn't know what to do with ReplicaSet.

@hasezoey can you review my changes? And please fix other parts of README otherwise users blow up with issues. Tnx.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ And one of those:
 
 ### Configuring which mongod binary to use
 
-The default behaviour is that version ~~`latest`~~`4.0.25` for your OS will be downloaded. By setting [Environment variables](https://nodkz.github.io/mongodb-memory-server/docs/api/config-options) you are able to specify which version and binary will be downloaded:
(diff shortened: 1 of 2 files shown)


hasezoey:
i actually am currently working on 7.2, where the error will be improved (adding StateError before AssertionError), and reworking readme

README.md
```diff
@@ -131,7 +125,7 @@ mongod.getInstanceInfo();
 All options are optional.
 
 ```js
-const mongod = new MongoMemoryServer({
+const mongod = await MongoMemoryServer.create({
```
hasezoey, line 128: for the options example-only, it is actually not wrong to use `new MMS()`

hasezoey:
merged as cbfa95e74606ba3d22ddb90277316541cef1e17a

PR #506, opened 2021-07-14
fix(errors): EACCES: do not throw on EACCES

bcmarinacci:
Hi! 👋 Thanks heaps for creating and maintaing such a useful library! 🙇

I noticed that when the Node.js process doesn't have permission to access the OS cache directory, an `EACCES` error is thrown and the process crashes. The following error happens to us in our CI environment:

```bash
EACCES: permission denied, stat '/root/.cache/mongodb-binaries/mongod-x64-ubuntu-4.0.25'
```

This pull request updates the handling of `EACCES` errors to match the way `ENOENT` errors are handled: returning `undefined` from `statPath()` without throwing an error. The result is that a MongoDB binary should be able to be installed to a local directory when access to the OS cache directory is restricted. The JSDoc comment was also updated and a test for this specific case was added.

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/__tests__/utils.test.ts
+++ b/packages/mongodb-memory-server-core/src/util/__tests__/utils.test.ts
@@ -31,7 +31,7 @@ describe('utils', () => {
   });
 
   describe('statPath', () => {
-    it('should throw an error if code is not "ENOENT"', async () => {
+    it('should throw an error if code is not "ENOENT" or "EACCES"', async () => {
       const retError = new Error();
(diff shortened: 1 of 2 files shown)


hasezoey (approved):

hasezoey:
thanks, didnt even think about `EACCES`, but it is good to treat it equal to an non-existing path (at least for this library and its use-cases)

hasezoey:
code looks good, could you fix the tests before being ready to merge?

bcmarinacci:
@hasezoey Thanks, mate! The tests should be ✅ now. 😊

hasezoey (approved):

PR #542, opened 2021-09-20
style(MongoInstance.ts): fix typo

SPodjasek:
Nothing much to say here, it just caught my eye. Thank for writing such a great library...


hasezoey (approved):
thanks for finding & fixing this typo

PR #574, opened 2021-11-05
fix(MongoMemoryServer): invert `isNew` for dbPaths

### LINKED ISSUE #573: Variable `isNew` is wrongly inverted when `dbPath` is defined and a new instance is created
I think this should be the other way round: A dbPath is to be considered new if there are no files in it (`files.length === 0`).

https://github.com/nodkz/mongodb-memory-server/blob/2a94fb0b3b382ecad2cb72e17968e6ffcfaeb9ee/packages/mongodb-memory-server-core/src/MongoMemoryServer.ts#L363-L365

Steditor:
This closes #573: a folder without any files indicates a new database.

## Related Issues

- #573


hasezoey (approved):
thanks for noticing and fixing this, i also added some tests that this does not happen in the future for this case again

PR #585, opened 2021-12-15
docs(docker): fix dockefile typo

sebaplaza:
This PR fixes a bad typo in documentation.

We write `dockerfile` instead of `dockefile`


hasezoey (approved):
thanks, didnt notice this one

PR #591, opened 2022-01-03
updated MongoMemoryServer options part

sabyr-cpu:
changed example code to up-to-date version, as the previous one returns error

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ await mongod.stop();
 All options are optional.
 
 ```js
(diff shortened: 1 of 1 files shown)


hasezoey:
> as the previous one returns error

what errors?

also, the example code is still correct as it currently is in master

sabyr-cpu:
oh, i am so sorry i was trying to get uri of MongoMemoryServer instance which was not started yet

PR #636, opened 2022-05-02
feat: add environment variable to use archive name for binary name

### LINKED ISSUE #609: Release is not part of the binary name identifier
## Versions

- NodeJS: 16.4.0 and 12.22.10
- mongodb-memory-server-*: 8.4.0 
- mongodb(the binary version): 5.0.3
- mongodb(the js package): 4.3.1
- mongoose: 6.2.3
- system: Linux

package: mongo-memory-server 

## What is the Problem?

I specified the `MONGOMS_DOWNLOAD_DIR` in my Jenkinsfile so that the tests could share the binaries between different runs and I encountered a problem when using different docker images of node:

- With docker.io/library/node:16.14.0 it downloads the following binary https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-debian10-5.0.3.tgz
- With docker.io/library/node:12.22.10 it downloads the following binary https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-debian92-5.0.3.tgz

However the output file name of both is just `mongod-x64-debian-5.0.3`, without the debian release version.
So when I try to run the test in the 12.22.10 container, after the binary has been downloaded in a 16.4.0 container, I get the following error: 

```txt
  MongoMS:MongoInstance Mongo[42073]: start: Starting Processes +8ms
  MongoMS:MongoInstance Mongo[42073]: _launchMongod: Launching Mongod Process +0ms
  MongoMS:MongoInstance Mongo[42073]: prepareCommandArgs +0ms
  MongoMS:MongoInstance Mongo[42073]: prepareCommandArgs: final argument array:["--port","42073","--dbpath","/tmp/mongo-mem--320-F3eEvdNstaky","--storageEngine","wiredTiger","--bind_ip","127.0.0.1","--noauth"] +0ms
  MongoMS:MongoInstance Mongo[42073]: _launchKiller: Launching Killer Process (parent: 320, child: 343) +21ms
  MongoMS:MongoInstance Mongo[42073]: stderrHandler: ""/shared_caches/mongodbms-binaries-cache/mongod-x64-debian-5.0.3: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.25' not found (required by /shared_caches/mongodbms-binaries-cache/mongod-x64-debian-5.0.3)
  MongoMS:MongoInstance /shared_caches/mongodbms-binaries-cache/mongod-x64-debian-5.0.3: /usr/lib/x86_64-linux-gnu/libcurl.so.4: version `CURL_OPENSSL_4' not found (required by /shared_caches/mongodbms-binaries-cache/mongod-x64-debian-5.0.3)
  MongoMS:MongoInstance /shared_caches/mongodbms-binaries-cache/mongod-x64-debian-5.0.3: /usr/lib/x86_64-linux-gnu/libssl.so.1.1: version `OPENSSL_1_1_1' not found (required by /shared_caches/mongodbms-binaries-cache/mongod-x64-debian-5.0.3)
  MongoMS:MongoInstance "" +25ms
  MongoMS:MongoInstance Mongo[42073]: closeHandler: Mongod instance closed with an non-0 (or non 12 on windows) code! +0ms
  MongoMS:MongoInstance Mongo[42073]: closeHandler: "1" "null" +0ms
  MongoMS:MongoMemoryServer Mongo[unknown]: stop: Called .stop() method +57ms
  MongoMS:MongoMemoryServer Mongo[unknown]: stop: "instanceInfo" is not defined (never ran?) +0ms
Error: Jest: Got error running globalSetup - /path-to-repo/test/e2e/setup.js, reason: Instance Exited before being ready and without throwing an error!
    at MongoInstance.<anonymous> (/path-to-repo/node_modules/mongodb-memory-server-core/lib/util/MongoInstance.js:140:25)
    at Object.onceWrapper (events.js:421:26)
    at MongoInstance.emit (events.js:314:20)
    at MongoInstance.closeHandler (/path-to-repo/node_modules/mongodb-memory-server-core/lib/util/MongoInstance.js:276:14)
    at ChildProcess.emit (events.js:314:20)
    at maybeClose (internal/child_process.js:1022:16)
    at Socket.<anonymous> (internal/child_process.js:444:11)
    at Socket.emit (events.js:314:20)
    at Pipe.<anonymous> (net.js:675:12)
error Command failed with exit code 1.
```

Yes, the problem is, that the release is not part of the binary name. The [getBinaryName](https://github.com/nodkz/mongodb-memory-server/blob/bbddd350626950dffa89592e740420d31f07df93/packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts#L216) function should also add the release to the binary name.

skrtheboss:
## Related Issues

- fixes #609

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
+++ b/packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
@@ -4,8 +4,9 @@ import { assertion, checkBinaryPermissions, isNullOrUndefined, pathExists } from
 import * as path from 'path';
 import { arch, homedir, platform } from 'os';
 import findCacheDir from 'find-cache-dir';
-import getOS, { AnyOS, isLinuxOS, OtherOS } from './getos';
+import { getOS, AnyOS, isLinuxOS, OtherOS } from './getos';
 import { NoRegexMatchError, NoSystemBinaryFoundError, ParseArchiveRegexError } from './errors';
+import { MongoBinaryDownloadUrl } from './MongoBinaryDownloadUrl';
 
 const log = debug('MongoMS:DryMongoBinary');
 
@@ -14,6 +15,7 @@ export interface BaseDryMongoBinaryOptions {
   downloadDir?: string;
   os?: AnyOS;
   arch?: string;
+  platform?: string;
   systemBinary?: string;
 }
 
@@ -24,6 +26,7 @@ export interface DryMongoBinaryOptions extends BaseDryMongoBinaryOptions {
 export interface DryMongoBinaryNameOptions {
   version: NonNullable<DryMongoBinaryOptions['version']>;
   arch: NonNullable<DryMongoBinaryOptions['arch']>;
+  platform: NonNullable<DryMongoBinaryOptions['version']>;
   os: NonNullable<DryMongoBinaryOptions['os']>;
 }
 
@@ -52,10 +55,6 @@ export class DryMongoBinary {
    * Binaries already found, values are: [Version, Path]
    */
   static binaryCache: Map<string, string> = new Map();
-  /**
-   * Cache the "getOS" call, so that not much has to be re-executed over and over
-   */
-  static cachedGetOs: AnyOS;
 
   /**
    * Try to locate an existing binary
@@ -111,15 +110,12 @@ export class DryMongoBinary {
       ? { version: defaultVersion }
       : opts;
 
-    if (isNullOrUndefined(this.cachedGetOs)) {
-      this.cachedGetOs = await getOS();
-    }
-
     const final: Required<DryMongoBinaryOptions> = {
       version: ensuredOpts.version || defaultVersion,
       downloadDir:
         resolveConfig(ResolveConfigVariables.DOWNLOAD_DIR) || ensuredOpts.downloadDir || '',
-      os: ensuredOpts.os ?? this.cachedGetOs,
+      os: ensuredOpts.os ?? (await getOS()),
+      platform: ensuredOpts.platform || platform(),
       arch: ensuredOpts.arch || arch(),
       systemBinary:
         resolveConfig(ResolveConfigVariables.SYSTEM_BINARY) || ensuredOpts.systemBinary || '',
@@ -208,12 +204,22 @@ export class DryMongoBinary {
    * Get the full path with filename
    * @returns Absoulte Path with FileName
    */
-  static getBinaryName(opts: DryMongoBinaryNameOptions): string {
+  static async getBinaryName(opts: DryMongoBinaryNameOptions): Promise<string> {
     log('getBinaryName');
-    const addExe = platform() === 'win32' ? '.exe' : '';
-    const dist = isLinuxOS(opts.os) ? opts.os.dist : opts.os.os;
 
-    return `mongod-${opts.arch}-${dist}-${opts.version}${addExe}`;
+    let binaryName: string;
+
+    if (envToBool(resolveConfig(ResolveConfigVariables.USE_ARCHIVE_NAME_FOR_BINARY_NAME))) {
+      const archiveName = await new MongoBinaryDownloadUrl(opts).getArchiveName();
+      binaryName = path.parse(archiveName).name;
+    } else {
+      const addExe = opts.platform === 'win32' ? '.exe' : '';
+      const dist = isLinuxOS(opts.os) ? opts.os.dist : opts.os.os;
+
+      binaryName = `mongod-${opts.arch}-${dist}-${opts.version}${addExe}`;
+    }
+
+    return binaryName;
   }
 
   /**
@@ -266,7 +272,7 @@ export class DryMongoBinary {
       relative: '',
       resolveConfig: '',
     };
-    const binaryName = this.getBinaryName(opts);
+    const binaryName = await this.getBinaryName(opts);
     // Assign "node_modules/.cache" to modulesCache
 
     // if we're in postinstall script, npm will set the cwd too deep
--- a/packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
@@ -1,9 +1,8 @@
-import { AnyOS, LinuxOS } from './getos';
+import { getOS, AnyOS, LinuxOS } from './getos';
 import { resolveConfig, ResolveConfigVariables } from './resolveConfig';
 import debug from 'debug';
 import * as semver from 'semver';
 import { isNullOrUndefined } from './utils';
-import { BaseDryMongoBinaryOptions, DryMongoBinary } from './DryMongoBinary';
 import { URL } from 'url';
 import {
   KnownVersionIncompatibilityError,
@@ -14,10 +13,10 @@ import {
 const log = debug('MongoMS:MongoBinaryDownloadUrl');
 
 export interface MongoBinaryDownloadUrlOpts {
-  version: NonNullable<BaseDryMongoBinaryOptions['version']>;
+  version: string;
   platform: string;
-  arch: NonNullable<BaseDryMongoBinaryOptions['arch']>;
-  os?: BaseDryMongoBinaryOptions['os'];
+  arch: string;
+  os?: AnyOS;
 }
 
 /**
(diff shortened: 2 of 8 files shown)


packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
```diff
@@ -14,10 +13,10 @@ import {
 const log = debug('MongoMS:MongoBinaryDownloadUrl');
 
 export interface MongoBinaryDownloadUrlOpts {
-  version: NonNullable<BaseDryMongoBinaryOptions['version']>;
+  version: string;
   platform: string;
-  arch: NonNullable<BaseDryMongoBinaryOptions['arch']>;
-  os?: BaseDryMongoBinaryOptions['os'];
+  arch: string;
+  os?: AnyOS;
```
hasezoey, line 19: why was this changed? the options were made this way to have a single place define the type across the project

packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
```diff
@@ -14,6 +15,7 @@ export interface BaseDryMongoBinaryOptions {
   downloadDir?: string;
   os?: AnyOS;
   arch?: string;
+  platform?: string;
```
hasezoey, line 18: See other comment about this, this message is just for reference

packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
```diff
@@ -111,15 +110,12 @@ export class DryMongoBinary {
...
-      this.cachedGetOs = await getOS();
-    }
-
     const final: Required<DryMongoBinaryOptions> = {
       version: ensuredOpts.version || defaultVersion,
       downloadDir:
         resolveConfig(ResolveConfigVariables.DOWNLOAD_DIR) || ensuredOpts.downloadDir || '',
-      os: ensuredOpts.os ?? this.cachedGetOs,
+      os: ensuredOpts.os ?? (await getOS()),
+      platform: ensuredOpts.platform || platform(),
```
hasezoey, line 118: See other comment about this, this message is just for reference

packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
```diff
@@ -24,6 +26,7 @@ export interface DryMongoBinaryOptions extends BaseDryMongoBinaryOptions {
 export interface DryMongoBinaryNameOptions {
   version: NonNullable<DryMongoBinaryOptions['version']>;
   arch: NonNullable<DryMongoBinaryOptions['arch']>;
+  platform: NonNullable<DryMongoBinaryOptions['version']>;
```
hasezoey, line 29: why exactly was this added? i currently dont quite see where this used that it would be required to be in here

packages/mongodb-memory-server-core/src/util/getos/index.ts
```diff
@@ -42,21 +42,28 @@ export function isLinuxOS(os: AnyOS): os is LinuxOS {
...
+    // Linux is a special case.
+    if (osName === 'linux') {
+      cachedOs = await getLinuxInformation();
+    } else {
+      cachedOs = { os: osName };
+    }
   }
 
-  return { os: osName };
+  return cachedOs;
```
hasezoey, line 64: why change the cache to be in getos instead of leaving it where it was?

packages/mongodb-memory-server-core/src/util/__tests__/dryBinary.test.ts
```diff
@@ -155,6 +157,46 @@ describe('DryBinary', () => {
     });
   });
 
+  describe('getBinaryName', () => {
+    afterEach(() => {
+      delete process.env[envName(ResolveConfigVariables.USE_ARCHIVE_NAME_FOR_BINARY_NAME)];
+    });
```
hasezoey, line 163: this should probably be also run in a `beforeAll` to ensure it starts at default

packages/mongodb-memory-server-core/src/util/__tests__/dryBinary.test.ts
```diff
@@ -155,6 +157,46 @@ describe('DryBinary', () => {
...
+
+    it('should return archive style ', async () => {
```
hasezoey, line 165:
```suggestion
    it('should return binary name matching archive name', async () => {
```

packages/mongodb-memory-server-core/src/util/__tests__/dryBinary.test.ts
```diff
@@ -155,6 +157,46 @@ describe('DryBinary', () => {
...
+            os: 'linux',
+            dist: 'Ubuntu Linux',
+            release: '20.04',
+          },
+          version: '5.0.3',
+        })
+      ).toEqual('mongodb-linux-x86_64-ubuntu2004-5.0.3');
+    });
+
+    it('should return non archive style ', async () => {
```
hasezoey, line 182:
```suggestion
    it('should return binary name not matching archive name ', async () => {
```

packages/mongodb-memory-server-core/src/util/__tests__/dryBinary.test.ts
```diff
@@ -155,6 +157,46 @@ describe('DryBinary', () => {
...
+  describe('getBinaryName', () => {
+    afterEach(() => {
+      delete process.env[envName(ResolveConfigVariables.USE_ARCHIVE_NAME_FOR_BINARY_NAME)];
+    });
+
+    it('should return archive style ', async () => {
+      process.env[envName(ResolveConfigVariables.USE_ARCHIVE_NAME_FOR_BINARY_NAME)] = 'true';
+
+      expect(
+        await DryMongoBinary.getBinaryName({
```
hasezoey, line 169: because `getos` is now a call in this function, it should probably have at least one spy attached to see that it is not actually called when providing `os`

packages/mongodb-memory-server-core/src/util/__tests__/dryBinary.test.ts
```diff
@@ -155,6 +157,46 @@ describe('DryBinary', () => {
     });
   });
 
+  describe('getBinaryName', () => {
```
hasezoey, line 160: maybe also add a test for a different architecture and different platform combination (eg arm, windows)

packages/mongodb-memory-server-core/src/util/resolveConfig.ts
```diff
@@ -23,6 +23,7 @@ export enum ResolveConfigVariables {
   RUNTIME_DOWNLOAD = 'RUNTIME_DOWNLOAD',
   USE_HTTP = 'USE_HTTP',
   SYSTEM_BINARY_VERSION_CHECK = 'SYSTEM_BINARY_VERSION_CHECK',
+  USE_ARCHIVE_NAME_FOR_BINARY_NAME = 'USE_ARCHIVE_NAME_FOR_BINARY_NAME',
```
hasezoey, line 26: from what i can tell, this option is currently missing from the docs, please add it to `docs/api/config-options` (in the style of the other options)

hasezoey (changes requested):
Thanks for putting up a pr for your issue, i have some questions and some suggestions (see below for the lines)

packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
skrtheboss: The problem is that there is a circular dependency between `DryMongoBinary.ts` and `MongoBinaryDownloadUrl.ts`.
Since  DryMongoBinary uses the getArchiveName from MongoBinaryDownloadUrl.

packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
skrtheboss: It is required in the DryMongoBinary's getBinaryName, `await new MongoBinaryDownloadUrl(opts).getArchiveName()`.

packages/mongodb-memory-server-core/src/util/getos/index.ts
skrtheboss: Since getOs is called in other files as well, I thought it would be more appropriate to cache it here.

packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
hasezoey: thanks for pointing this out, i will think about refactoring these types at some point

hasezoey (approved):
This PR looks good to me now, i will for now merge it into beta, i dont know when this will be released on master (though a beta version will be available)

PR #660, opened 2022-06-11
chore: correct MongoMemoryReplSet example

idanavr:


hasezoey:
Why should this be changed back? The example code is valid since at least 7.0.0

PR #691, opened 2022-09-05
feat: upgrade workflow for node 18

shakaran:
Remove support node 12 for version 9.0


hasezoey:
Thanks for the PR, but i think i will at least put it off for now, because even though node 12 is EOL, some projects (that maintain / help maintain) still actively support node 12 and will likely not switch soon, also it is not much of a burden

btw, there are some places in the code where there is special code depending on the node version (like node changing which `fs.remove` functions with `recursive` to use)

PR #716, opened 2022-11-14
fix(MongoMemoryReplSet): pass-through option "launchTimeout" correctly

camspiers:
The launchTimeout option if specified in instanceOpts for MongoMemoryReplSet doesn't get passed into the eventual options for each underlying MongoMemoryServer. This change modifies the base opts type to include launchTimeout as a common instance option. 

I have run the automated tests, but I am not sure exactly what else needs testing.

### DIFF
--- a/packages/mongodb-memory-server-core/src/MongoMemoryReplSet.ts
+++ b/packages/mongodb-memory-server-core/src/MongoMemoryReplSet.ts
@@ -322,6 +322,9 @@ export class MongoMemoryReplSet extends EventEmitter implements ManagerAdvanced 
     if (baseOpts.replicaMemberConfig) {
       opts.replicaMemberConfig = baseOpts.replicaMemberConfig;
     }
+    if (baseOpts.launchTimeout) {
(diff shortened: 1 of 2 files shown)


hasezoey:
thanks for fixing this, i really need to refactor the way options are handled - could you also add some tests or should i do it?

simllll:
We have some flaky tests right now with replset: 2 .. where we get      MongoBulkWriteError: Server selection timed out after 30000 ms
can this be related?

hasezoey:
@simllll please open a new issue

aside from that, no `launchTimeout` is a timeout for the case where the `mongod`-binary child-process does not respond with any events or errors

if you use replsets, then sadly this is a problem of how mongodb just works, it may just decide at first it has a primary (which MMS will listen and wait for until ready), but it may just soon after decide it steps-down and so has no primary for some time - which fails all connections wanting a primary, there is nothing i know of that MMS can do about

hasezoey:
i take the staleness as that i should add the tests

PR #728, opened 2023-01-06
Widen dependency range for mongodb

dobesv:
The current dependency only allows mongodb >=4.11.0  and <4.12.0 but 4.12 is out and should be compatible.

When libraries restrict the dependency ranges it can cause issues beyond simply having multiple copies of the library installed, but also instrumentation / monitoring libraries might instrument the "wrong" copy of mongodb, as is the case for us.

### DIFF
--- a/packages/mongodb-memory-server-core/package.json
+++ b/packages/mongodb-memory-server-core/package.json
@@ -50,7 +50,7 @@
     "get-port": "^5.1.1",
     "https-proxy-agent": "^5.0.1",
     "md5-file": "^5.0.0",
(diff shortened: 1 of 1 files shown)


hasezoey:
merged as 605614db338a0e7c667dcfbb38439cca337277cd

dobesv:
> this shouldnt really affect anything outside of MMS

I do agree it *shouldn't* affect anything outside of MMS but it unfortunately did affect us.

It broke our opentelemetry metrics because for some reason opentelemetry ended up picking the version from mongodb-memory-server to insttrument instead of the one used in the main app.

Issue #735, opened 2023-01-19
Add Support for "elementaryOS 6.1"


chtpl:
## Versions

- Wanted System: `elementaryOS 6.1`
- Architecture: `x86_64`
- mongodb-memory-server: `8.11.2`
- node: `14.21.2`
- npm: `6.14.17`

## Detection

```sh
$ cat /etc/upstream-release/lsb-release

ID=Ubuntu
VERSION_ID=20.04
VERSION_CODENAME=focal
PRETTY_NAME="Ubuntu 20.04.5 LTS"

$ cat /etc/os-release

NAME="elementary OS"
VERSION="6.1 Jólnir"
ID=elementary
ID_LIKE=ubuntu
PRETTY_NAME="elementary OS 6.1 Jólnir"
LOGO=distributor-logo
VERSION_ID="6.1"
HOME_URL="https://elementary.io/"
DOCUMENTATION_URL="https://elementary.io/docs/learning-the-basics"
SUPPORT_URL="https://elementary.io/support"
BUG_REPORT_URL="https://github.com/elementary/triage/issues/new"
PRIVACY_POLICY_URL="https://elementary.io/privacy-policy"
VERSION_CODENAME=jolnir
UBUNTU_CODENAME=focal

$ cat /usr/lib/os-release

NAME="elementary OS"
VERSION="6.1 Jólnir"
ID=elementary
ID_LIKE=ubuntu
PRETTY_NAME="elementary OS 6.1 Jólnir"
LOGO=distributor-logo
VERSION_ID="6.1"
HOME_URL="https://elementary.io/"
DOCUMENTATION_URL="https://elementary.io/docs/learning-the-basics"
SUPPORT_URL="https://elementary.io/support"
BUG_REPORT_URL="https://github.com/elementary/triage/issues/new"
PRIVACY_POLICY_URL="https://elementary.io/privacy-policy"
VERSION_CODENAME=jolnir
UBUNTU_CODENAME=focal

$ cat /etc/lsb-release

DISTRIB_ID=elementary
DISTRIB_RELEASE=6.1
DISTRIB_CODENAME=jolnir
DISTRIB_DESCRIPTION="elementary OS 6 Jólnir"
```

hasezoey:
does elementaryOS 6.1 currently not work? because elementaryOS 6 should already be handled (by either explicit or implicit upstream)

chtpl:
unfortunately it does not work at the moment

hasezoey:
[...]

PR #736, opened 2023-01-26
feat(getos): allow to override the release file

tintou:
The lsb_release program upstream allows to use the LSB_OS_RELEASE environment variable to override the release file to consider, it would be nice to support this here.

For instance, we plan to move away from /etc/upstream-release/lsb-release in elementary OS to /usr/lib/upstream-os-release

Note that the lsb-release file (and its format) is now legacy and isn't even used by lsb_release itself.

## Related Issues

- #735

Note that this is my first time dealing with TypeScript so any feedback welcome :slightly_smiling_face:

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/getos/index.ts
+++ b/packages/mongodb-memory-server-core/src/util/getos/index.ts
@@ -70,10 +70,17 @@ export async function getOS(): Promise<AnyOS> {
 /** Function to outsource Linux Information Parsing */
 async function getLinuxInformation(): Promise<LinuxOS> {
   // Structure of this function:
-  // 1. get upstream release, if possible
-  // 2. get os release (etc) because it has an "id_like"
-  // 3. get os release (usr) because it has an "id_like"
-  // 4. get lsb-release (etc) as fallback
+  // 1. get os release from environment variable, if set
+  // 2. get upstream release, if possible
+  // 3. get os release (etc) because it has an "id_like"
+  // 4. get os release (usr) because it has an "id_like"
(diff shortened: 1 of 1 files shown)


hasezoey:
> Note that this is my first time dealing with TypeScript so any feedback welcome slightly_smiling_face

yes it is the correct way, only problems are:
- no tests
- no documentation
- passing in a possibly undefined value into a function expecting a string

if you want i can make the necessary changes, but would like the question of documentation for the `LSB_OS_RELEASE` env variable resolved first (to see it is not distro specific)

also as a side-note, is the `LSB_OS_RELEASE` always a os-release style file or can it also be a lsb-release file?

> allows to use the LSB_OS_RELEASE environment variable to override the release file to consider

do you have some documentation regarding this as a standard? because from my searching i could not find anything related and on the distros i have tried i never saw this environment variable

> Note that the lsb-release file (and its format) is now legacy and isn't even used by lsb_release itself.

i know that it is legacy, but as long as it exists, it provides all the information that is required for here (distro name, distro version)
also as a note, the command itself is not used because i personally found that it is not provided as common as the files are, but the files are always existing (most notably docker images)

tintou:
Sorry, I'm reconsidering this, I had to use LSB_OS_RELEASE in a very specific use case when I needed it to return the upstream distro codename but this shouldn't ever be the case here.

As a side note:
 * `LSB_OS_RELEASE` is guaranteed to be an os-release format
 * `LSB_ETC_LSB_RELEASE` used to exist for the lsb-release format

But it probably doesn't make sense to add it here.

tintou:
Thank you for the kind review though!

PR #762, opened 2023-04-03
download correct package for arch

jeysal:
Arch Linux is on OpenSSL 3 (instead of 1.1), like Ubuntu 22.04 (not like 20.04), so currently the `mongod` startup fails because of missing `libcrypto.so.1.1` (unless one has the `openssl-1.1` package installed for this or some other reason).

Tested by making this same change locally in the distributed `node_modules/mongodb-memory-server-core`.

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts
+++ b/packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts
@@ -1684,7 +1684,7 @@ describe('MongoBinaryDownloadUrl', () => {
   });
 
   describe('getLinuxOSVersionString()', () => {
-    it('should give an warning about "alpine"', () => {
+    it('should give a warning about "alpine"', () => {
       jest.spyOn(console, 'warn').mockImplementation(() => void 0);
       const du = new MongoBinaryDownloadUrl({
         platform: 'linux',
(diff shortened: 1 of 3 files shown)


hasezoey:
Looks good to me, could you add tests for this and check if the documentation regarding arch are correct?

jeysal:
I added a test case similar to those already present, but I don't know how correct e.g. the `ubuntu1604` string is since the test cases test some internal functions that are pretty far from the user-facing things that happen. Feel free to adapt.
I didn't see relevant documentation that mentions anything Arch-specific, but could add a changelog entry to mention that this is now working for the current Arch Linux with OpenSSL 3 and not old ones with OpenSSL 1.1

packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts
```diff
@@ -1723,6 +1723,24 @@ describe('MongoBinaryDownloadUrl', () => {
       expect(du.getLegacyVersionString).toHaveBeenCalledTimes(1);
       expect(ret).toBe('');
     });
+
+    it('should give a warning about "arch"', () => {
```
hasezoey, line 1727: that is actually the wrong place, this `describe` block tests the that `getLinuxOSVersionString` behaves correctly, distro specific tests are in `MongoBinaryDownloadUrl -> getDownloadUrl() -> for linux` and there already exist similar tests for arch called `for arch` and `for unpopular arch`, though i would probably later move them into their own describe block

jeysal:
Feel free to make the changes and ship, otherwise I can can get to it tomorrow, am out and only on my phone atm

hasezoey:
added commits to add tests (and move tests), update docs and [also updated your commit message](https://github.com/nodkz/mongodb-memory-server/blob/master/.github/CONTRIBUTING.md#commit-structure) (because i dont plan on squashing this PR)

are you fine with the changes or does anything need to be changed?

jeysal:
@hasezoey looks reasonable :) thanks

hasezoey (approved):

PR #787, opened 2023-08-16
fix: Fix binary download on debian testing/unstable

orgads:
VERSION_ID was removed. See
https://tracker.debian.org/news/1433360/accepted-base-files-13-source-into-unstable/

and a similar bug:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008735

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
@@ -289,7 +289,9 @@ export class MongoBinaryDownloadUrl implements MongoBinaryDownloadUrlOpts {
       throw new UnknownVersionError(this.version);
     }
 
-    if (release >= 11 || ['unstable', 'testing'].includes(os.release)) {
+    const isTesting = ['unstable', 'testing', ''].includes(os.release);
+
+    if (isTesting || release >= 11) {
       // Debian 11 is compatible with the binaries for debian 10
       // but does not have binaries for before 5.0.8
       // and only set to use "debian10" if the requested version is not a latest version
@@ -309,10 +311,10 @@ export class MongoBinaryDownloadUrl implements MongoBinaryDownloadUrlOpts {
       name += '71';
     }
 
-    if (release >= 10) {
+    if (isTesting || release >= 10) {
       if (semver.lt(coercedVersion, '4.2.1') && !testVersionIsLatest(this.version)) {
         throw new KnownVersionIncompatibilityError(
-          `Debian ${release}`,
+          `Debian ${release || os.release || os.codename}`,
           this.version,
           '>=4.2.1',
           'Mongodb does not provide binaries for versions before 4.2.1 for Debian 10+ and also cannot be mapped to a previous Debian release'
--- a/packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts
+++ b/packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts
@@ -554,6 +554,22 @@ describe('MongoBinaryDownloadUrl', () => {
           );
         });
 
+        it('for debian testing/unstable', async () => {
+          const du = new MongoBinaryDownloadUrl({
+            platform: 'linux',
+            arch: 'x64',
+            version: '5.0.9',
+            os: {
+              os: 'linux',
+              dist: 'debian',
+              release: '',
+            },
+          });
+          expect(await du.getDownloadUrl()).toBe(
+            'https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-debian11-5.0.9.tgz'
+          );
+        });
+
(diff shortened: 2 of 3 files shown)


hasezoey:
what exactly has this PR to do with the mentioned email thread?
and what exactly does this fix?

please add a test to illustrate what the change is meant to cover. [tests can be found here](https://github.com/nodkz/mongodb-memory-server/blob/master/packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts#L428)

orgads:
Updated the commit message, added tests and fixed some more cases.

packages/mongodb-memory-server-core/src/util/__tests__/MongoInstance.test.ts
```diff
@@ -209,13 +210,17 @@ describe('MongodbInstance', () => {
...
+
+      if (!isDebian10OrHigher(os)) {
+        const gotPort = await getPort({ port: 27445 });
+        const mongod = await MongodbInstance.create({
+          instance: { port: gotPort, dbPath: tmpDir },
+          binary: { version: '4.0.28' }, // explicit version instead of default to not mess it up later
+        });
+        expect(mongod.mongodProcess!.pid).toBeGreaterThan(0);
+        await mongod.stop();
+      }
```
hasezoey, line 223: this is a change that should probably more properly be handled and across all those tests for platforms where it is not available

packages/mongodb-memory-server-core/src/util/getos/index.ts
```diff
@@ -131,6 +131,14 @@ export function isValidOs(os: LinuxOS | undefined): os is LinuxOS {
...
 }
 
+export function isDebian10OrHigher(os: AnyOS): boolean {
+  if (!isLinuxOS(os) || os.dist !== 'debian') {
+    return false;
+  }
+
+  return ['unstable', 'testing', ''].includes(os.release) || parseFloat(os.release) >= 10;
+}
+
```
hasezoey, line 141: i dont think this is the proper place for this, `getos` is intended to get all information for a given system, but not modify it or distro specific functions, i think the better place would be in `MongoBinaryDownloadUrl` or `utils`

packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
```diff
@@ -289,7 +289,9 @@ export class MongoBinaryDownloadUrl implements MongoBinaryDownloadUrlOpts {
       throw new UnknownVersionError(this.version);
     }
 
-    if (release >= 11 || ['unstable', 'testing'].includes(os.release)) {
+    const isTesting = ['unstable', 'testing', ''].includes(os.release);
```
hasezoey, line 292: please combine this with the one from the function, to have less duplication

packages/mongodb-memory-server-core/src/__tests__/testUtils/globalSetup.ts
```diff
@@ -1,13 +1,17 @@
...
 
 export = async function globalSetup(): Promise<void> {
   const defaultVersion = resolveConfig(ResolveConfigVariables.VERSION);
   assertion(!isNullOrUndefined(defaultVersion), new Error('Default version is not defined'));
   const versions = [defaultVersion, '4.0.28', '4.2.23', '4.4.22', '5.0.19', '6.0.6'];
+  const isNewDebian = isDebian10OrHigher(await getOS());
   // Ensure all required versions are downloaded for tests
   for (const version of versions) {
-    await MongoBinary.getPath({ version });
+    if (!isNewDebian || parseFloat(version) >= 4.2) {
```
hasezoey, line 13: instead of float, i would prefer having it use semver

hasezoey (changes requested):
mostly looks good to me, with some small changes

packages/mongodb-memory-server-core/src/util/__tests__/MongoInstance.test.ts
orgads: Removed. Will do this in a separate PR.

orgads: Actually I prefer to leave this to you. I don't have other platforms to test on.

packages/mongodb-memory-server-core/src/util/getos/index.ts
orgads: Removed.

packages/mongodb-memory-server-core/src/__tests__/testUtils/globalSetup.ts
orgads: Removed

packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts
orgads: I removed the function.

hasezoey (approved):

PR #811, opened 2023-10-12
fix(kill): should not try to open a connection to a not running ReplSet

arthursimas1:
I've a situation where I listen to kill signals and execute `.stop()` when needed. The replica set has already been killed because the `Ctrl + C` signal is propagated to the whole process group, then I waits forever to connect and send a graceful shutdown.


hasezoey (approved):
Looks good, though we should probably also add some options to the connection (like `connectTimeoutMS` which defaults to 30s)

PR #813, opened 2023-10-12
fix(kill): should not wait too much to open a connection to ReplSet

arthursimas1:


packages/mongodb-memory-server-core/src/util/__tests__/MongoInstance.test.ts
```diff
@@ -292,12 +292,30 @@ describe('MongodbInstance', () => {
...
+    const mongod = new MongodbInstance({
+      instance: {
+        replSet: 'testset',
+        ip: '127.0.0.1',
+        port: gotPort,
+        dbPath: tmpDir,
+      },
+    });
+    await mongod.start();
+    jest.spyOn(MongoClient, 'connect');
```
hasezoey, line 314: you spy on the function, but dont check if it was called

packages/mongodb-memory-server-core/src/util/__tests__/MongoInstance.test.ts
```diff
@@ -292,12 +292,30 @@ describe('MongodbInstance', () => {
...
-    await new Promise<void>((resolve) => setTimeout(resolve, 100));
+    while (dbUtil.isAlive(mongod.mongodProcess!.pid)) {
+      await new Promise<void>((resolve) => setTimeout(resolve, 5));
+    }
     await mongod.stop();
 
     expect(MongoClient.connect).not.toBeCalled();
   });
 
+  test('"kill" should not wait too much to open a connection to ReplSet', async () => {
```
hasezoey, line 303:
```suggestion
  test('"kill" should not wait too much to open a connection', async () => {
```

from what i can tell, this is not replset specific
(the other PR also added a test with this in the title, could you remove it?)

hasezoey (changes requested):
once the connection spy is actually checked, this PR would be good

packages/mongodb-memory-server-core/src/util/__tests__/MongoInstance.test.ts
arthursimas1: There's a condition `if (this.isReplSet)` to call `MongoClient.connect`. That's why I've written it in this way. I don't really know why its's done this way.

hasezoey: > There's a condition if (this.isReplSet) to call MongoClient.connect
i am sorry, you are correct. i will re-review this again tomorrow when i am more awake again.

> I don't really know why its's done this way.

if i remember correctly, this is because if mongod is running as a replset and receives a signal to stop (like SIGTERM), it will have a ~30 second timeout before actually shutting down (trying to reach another server to synchronize to), but this does not happen if a `shutdown` with option `force: 1` command is given

packages/mongodb-memory-server-core/src/util/MongoInstance.ts
```diff
@@ -445,6 +450,7 @@ export class MongoInstance extends EventEmitter implements ManagerBase {
             con = await MongoClient.connect(uriTemplate(ip, port, 'admin'), {
               ...this.extraConnectionOptions,
               directConnection: true,
+              serverSelectionTimeoutMS: 500,
```
hasezoey, line 453:
```suggestion
              serverSelectionTimeoutMS: 1000, // 1 second
```

just to be safe, it will be 1 second

packages/mongodb-memory-server-core/src/util/__tests__/MongoInstance.test.ts
```diff
@@ -279,7 +279,7 @@ describe('MongodbInstance', () => {
     expect(dbUtil.killProcess).not.toBeCalled();
   });
 
-  it('"kill" should not try to open a connection to a not running ReplSet', async () => {
+  it('"kill" should not try to open a connection to a not running instance', async () => {
```
hasezoey, line 282:
```suggestion
  it('"kill" should not try to open a connection to a not running ReplSet instance', async () => {
```

you were correct

packages/mongodb-memory-server-core/src/util/__tests__/MongoInstance.test.ts
```diff
@@ -292,12 +292,31 @@ describe('MongodbInstance', () => {
...
-    await new Promise<void>((resolve) => setTimeout(resolve, 100));
+    while (dbUtil.isAlive(mongod.mongodProcess!.pid)) {
+      await new Promise<void>((resolve) => setTimeout(resolve, 5));
+    }
     await mongod.stop();
 
     expect(MongoClient.connect).not.toBeCalled();
   });
 
+  test('"kill" should not wait too much to open a connection', async () => {
```
hasezoey, line 303:
```suggestion
  test('"kill" should not wait too much to open a connection to a ReplSet instance', async () => {
```

packages/mongodb-memory-server-core/src/util/__tests__/MongoInstance.test.ts
```diff
@@ -292,12 +292,31 @@ describe('MongodbInstance', () => {
...
+        port: gotPort,
+        dbPath: tmpDir,
+      },
+    });
+    await mongod.start();
+    jest.spyOn(MongoClient, 'connect');
+    process.kill(mongod.mongodProcess!.pid, 'SIGKILL');
+    await mongod.stop();
+    expect(MongoClient.connect).toBeCalled();
+  }, 8000); // the default serverSelectionTimeoutMS is 10s, so waiting for 8s seems ok to catch it
```
hasezoey, line 318: > the default serverSelectionTimeoutMS is 10s, so waiting for 8s seems ok to catch it

isnt the default 30 seconds?
https://github.com/mongodb/node-mongodb-native/blob/2ab2189561736e354437df88d2ea26cf35b483d1/src/connection_string.ts#L1078-L1081

hasezoey (changes requested):
Code looks good, just some semantics (some of which i wrongly suggested previously)

hasezoey:
as this PR had no more activity, but the changes are still good, i will modify it myself and then merge

packages/mongodb-memory-server-core/src/util/__tests__/MongoInstance.test.ts
hasezoey:
```suggestion
  }, 8000); // the default serverSelectionTimeoutMS is 30s, so waiting for 8s seems ok to catch it
```

PR #826, opened 2023-11-02
Add support for amazon linux 2023

### LINKED ISSUE #825: Amazon EC2 instance download failed, combination does not exist (403)
## Versions
- NodeJS: 16.20.1
- mongodb-memory-server-*: 9.0.1 
- mongodb(the binary version): 7.0.0
- mongodb(the js package): 6.2.0
- system: Linux in an ec2 instance

package: mongo-memory-server 

## What is your question?

Using mongo-memory-server to write unit tests and it works locally on my machine. The above versions are those of the ec2 instance. When sent to an ec2 container to build and run unit tests, for some reason it doesn't seem like it can fetch for the mongo binaries. I get the following error message:
```
Starting the MongoMemoryServer Instance failed, enable debug log for more information. Error:
DownloadError: Download failed for url "https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-amazon-7.0.0.tgz", Details: Status Code is 403 (MongoDB's 404). This means that the requested version-platform combination doesn't exist
Try to use different version 'new MongoMemoryServer({ binary: { version: 'X.Y.Z' } })'
List of available versions can be found here: https://www.mongodb.com/download-center/community/releases/archive
```
current config within my package.json: 
```
 "config": {
    "mongodbMemoryServer": {
      "version": "7.0.0",
      "debug": "on"
    }
  }
```

georgehb:
The os string in the download url for amazon linux 2023 should be "amazon2023"

## Related Issues

- fixes #825

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,7 @@ jobs:
           ref: ${{ github.event.inputs.git-ref }}
           fetch-depth: 5 # because this is an manual trigger, only fetch the latest 5
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
       - name: Load MongoDB binary cache
@@ -59,7 +59,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Use Node.js 18
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: 18.x
       - name: Install node_modules
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -24,7 +24,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Use Node.js 18
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: 18.x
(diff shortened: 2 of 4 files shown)


hasezoey (changes requested):
[please add tests](https://github.com/nodkz/mongodb-memory-server/blob/9288cf05ca7ce184c687301b48d724edcf972d67/packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownloadUrl.test.ts#L1021), at least for version `7.0.0` which seems to be supported for 2023

also how would versions like 6.0.6 be handled? because from what i can tell there are only releases for `amazon2` for the 6.0.x series, are they compatible with the `2023` version?

georgehb:
> also how would versions like 6.0.6 be handled? because from what i can tell there are only releases for `amazon2` for the 6.0.x series, are they compatible with the `2023` version?

According to their docs, mongodb doesn't support mongodb version less than 7 on amazon 2023
https://www.mongodb.com/docs/manual/administration/production-notes/#platform-support-matrix

I just tried it to see what would happen if you attempted to start mongod 6.0.6 on al2023 and at least on the base al2023 image, the mongo instance failed to start due to a missing library. So I'd say they aren't compatible

georgehb:
I have added a test

hasezoey (approved):
looks good to me, i will change the base branch to `beta` and merge it into there

Issue #827, opened 2023-11-03
"getport" sometimes throwing an error


tarasromil:
## Versions

- NodeJS: 16.20.2
- mongodb-memory-server-*: 9.0.1
- mongodb(the binary version): 6.0.10
- mongodb(the js package): 5.9.0
- system: MacOS 14.1

package: mongo-memory-server

## What is the Problem?

When I'm trun to run test using "vitest@0.34.6" in concurrency mode I have an error that port already in use. No always, but pretty often

```
Starting the MongoMemoryServer Instance failed, enable debug log for more information. Error:
 StdoutInstanceError: Port "15490" already in use
```

Want to notice that there is no error on **v8.16.0**

## Code Example

```ts
globalThis.__MONGO_DB__ = await MongoMemoryServer.create();
```

## Debug Output

<details>

```txt
stderr | unknown test
Starting the MongoMemoryServer Instance failed, enable debug log for more information. Error:
 StdoutInstanceError: Port "52929" already in use
    at MongoInstance.checkErrorInLine (/.../node_modules/.pnpm/mongodb-memory-server-core@9.0.1/node_modules/mongodb-memory-server-core/src/util/MongoInstance.ts:641:9)
    at MongoInstance.stdoutHandler (/.../node_modules/.pnpm/mongodb-memory-server-core@9.0.1/node_modules/mongodb-memory-server-core/src/util/MongoInstance.ts:615:10)
    at Socket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:315:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at Socket.Readable.push (node:internal/streams/readable:228:10)
    at Pipe.onStreamRead (node:internal/stream_base_commons:190:23)
```

</details>

## Do you know *why* it happenes?

Looks like there is a race condition after replacing "get-port" to own implementation

hasezoey:
i dont know how `vitest`'s concurrency works or how / where you create your server.
i also dont know why the new implementation would fail when the old one would not in the same scenario.

port acquisition and keeping is a fickle process, because you cant just "give me a free port and keep it on-hold until used" and also cannot be marked as "used for later", like it is tested that it is free with a small http server, which is closed to free the port so that it can be used by mongod, but in that small time another process may just get that port

AndreasGalster2:
[...]

PR #844, opened 2024-01-11
fix(getport): Randomize first port using crypto

noseworthy:
The current implementation of `getPort()` relies on using `Date.now()`to get the first port to try to launch mongo on, even when the `EXP_NET0LISTEN` flag is set. This causes a couple issues:
- If the `Date` module is mocked, it can result in `getFreePort()` returning the same first port every time.
- If multiple mongos are being started at once in parallel, it's possible for the same first port to be picked leading to port conflicts

In order to better randomize the initial port selection so that port conflicts can be avoided, instead of using `Date.now()` for an initial seed, use `crypto.randomInt()` which should provide more randomness and hopefully avoid some of these race conditions.

## Related Issues

- #827

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/getport/index.ts
+++ b/packages/mongodb-memory-server-core/src/util/getport/index.ts
@@ -50,7 +50,7 @@ export async function getFreePort(
   max_tries: number = MAX_DEFAULT_TRIES
 ): Promise<number> {
   // Get a random value from crypto to use as first port if none is given
-  firstPort = firstPort || validPort(crypto.randomInt(MIN_PORT, MAX_PORT));
+  firstPort = firstPort || validPort(crypto.randomInt(MIN_PORT, MAX_PORT + 1));
 
   // clear ports cache after some time, but not on an interval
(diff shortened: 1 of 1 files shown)


packages/mongodb-memory-server-core/src/util/getport/index.ts
```diff
@@ -48,10 +49,10 @@ export async function getFreePort(
   firstPort?: number,
   max_tries: number = MAX_DEFAULT_TRIES
 ): Promise<number> {
-  // use "Date" as a semi-random value to lessen conflicts between simultaneous tests
-  firstPort = firstPort || validPort(Date.now());
+  // Get a random value from crypto to use as first port if none is given
+  firstPort = firstPort || validPort(crypto.randomInt(MIN_PORT, MAX_PORT));
```
hasezoey, line 53:
```suggestion
  firstPort = firstPort || validPort(crypto.randomInt(MIN_PORT, MAX_PORT + 1));
```

the documentation says that parameter `max` is exclusive, but `MAX_PORT` is inclusive (a number that should be a valid port)

hasezoey (approved):
Thanks, didnt know about the crypto function (though i knew about Math.random, but for some reason i cant remember didnt use it)

Issue #797, opened 2023-09-11
Can't launch on Debian 12 (bookworm) : missing or cannot be opened: "libcrypto.so.1.1"


BenjD90:
## Versions

- NodeJS: 16.20.2
- mongodb-memory-server-*:  8.15.1
- mongodb(the binary version): 6.0.9
- mongodb(the js package): 4.17.1 (the one used by mongodb-memory-server)
- system: Linux, in a docker image based on [node:16.20.2-bookworm](https://hub.docker.com/layers/library/node/16.20.2-bookworm/images/sha256-6b451f951f78547bbe248f93b5aef0b7082a4a746747cad75527410fd5aa2c5c?context=explore)
package: mongo-memory-server

## What is the Problem?

The mongodb can't be launched : 
```
Instance failed to start because a library is missing or cannot be opened: "libcrypto.so.1.1"
```

## Code Example

```ts
await MongoMemoryServer.create({
	binary: {
		version: '6.0.9',
	},
});
```

## Do you know *why* it happenes?

This seems related to this issue :shrug:  https://github.com/nodejs/docker-node/issues/1915

Thanks in advance for any idea how I can fix this, may-be the same way [it is done with Fedora](https://github.com/nodkz/mongodb-memory-server/blob/master/packages/mongodb-memory-server-core/src/util/MongoBinaryDownloadUrl.ts#L334) ?

hasezoey:
there are currently no binaries directly for debian 12, highest mongodb builds for is debian 11, i dont know which version changed openssl (i think debian 11?), so your options are:
- manually install libcrypto1.1 (like for ubuntu #732) 
- use a lower debian version (like 11)
- try to use the ubuntu binary (via config option [`DISTRO`](https://nodkz.github.io/mongodb-memory-server/docs/api/config-options#distro))

BenjD90:
Thanks, I think I'll have to switch to MongoDB 7 if I want to keep using Debian 12 : https://www.mongodb.com/docs/v6.0/administration/production-notes/#platform-support-matrix

![image](https://github.com/nodkz/mongodb-memory-server/assets/8216338/2a67afd9-aa79-42ed-92d2-fc4143b45631)

hasezoey:
debian 12 is now supported, thanks to #858, released in [9.1.8](https://github.com/nodkz/mongodb-memory-server/releases/tag/v9.1.8)

[...]

PR #858, opened 2024-03-28
fix(MongoBinaryDownloadUrl): support Debian 12 (Bookworm)

pnordhus:
Binaries for Debian 12 are available for MongoDB >= 7.0.3.

See https://jira.mongodb.org/browse/SERVER-77231

## Related Issues

- #797


hasezoey (changes requested):
Thanks for doing those changes, the only thing missing is to [update the docs](https://github.com/nodkz/mongodb-memory-server/blob/master/docs/guides/supported-systems.md#debian).
Do you want to do this, or should i do it after the merge?

pnordhus:
@hasezoey Good point! I think it makes sense to include the doc update in this PR. Updated :relaxed:

hasezoey (approved):

PR #866, opened 2024-04-21
Update test-runners.md

Darkmift:
- fixed doc to reflect proper API (current example caused crashes for me)

## Related Issues

- #1


hasezoey:
> current example caused crashes for me

are you running ESM or CommonJS?

Darkmift:
> > current example caused crashes for me
> 
> are you running ESM or CommonJS?

ESM

hasezoey:
the example works and has been tested in CommonJS, in ESM it likely does not work as mongoose does not export `connection` there (as ESM does not allow changing the export once set)

hasezoey (approved):

PR #936, opened 2025-07-14
Http download retry issue 929 3

firefoxNX:
Added more retryable error codes.

### DIFF
--- a/packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
+++ b/packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
@@ -19,8 +19,20 @@ import { RequestOptions } from 'https';
 
 const log = debug('MongoMS:MongoBinaryDownload');
 
-const retryableStatusCodes = [503, 500];
-const retryableErrorCodes = ['ECONNRESET', 'ETIMEDOUT', 'ENOTFOUND', 'ECONNREFUSED'];
+const retryableStatusCodes = [429, 500, 503];
+
+const retryableErrorCodes = [
+  'ECONNRESET',
+  'ETIMEDOUT',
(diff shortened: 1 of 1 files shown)


packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
```diff
@@ -19,8 +19,20 @@ import { RequestOptions } from 'https';
...
+  'ETIMEDOUT',
+  'ENOTFOUND',
+  'ECONNREFUSED',
+  'EPIPE',
+  'EHOSTUNREACH',
+  'EAI_AGAIN',
+  'ENETUNREACH',
+  'ECONNABORTED',
+  'aborted',
+];
```
hasezoey, line 35: I think it would be a good idea to add some comments to non-self-explaining errors, for example what is `EAI_AGAIN`? (i did not even know about this error before looking it up)

hasezoey (approved):
Looks good to me, though some documentation for at least some errors would be great.
Also did you already test it in your CI with those changes?

firefoxNX:
> Looks good to me, though some documentation for at least some errors would be great. Also did you already test it in your CI with those changes?

No. I have not tested with these retryable codes in Jenkins CI (running inside docker)

We use artifactory which takes 1 hour to reflect the public npmjs changes. I can test with 10.2.0-beta.4 once that is available.

packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
firefoxNX: Yes there are lots of such error codes - https://nodejs.org/api/errors.html#nodejs-error-codes

PR #937, opened 2025-07-20
docs(guides): add vitest integration example

danielpza:
Added [vitest](https://vitest.dev) integration example

Also linked to the https://github.com/danielpza/vitest-mms package which aims to make it easier to setup up mongodb-memory-server with vitest (DISCLAIMER: I'm the author of the package, I can drop it if that's an issue)

### DIFF
--- a/docs/guides/integration-examples/test-runners.md
+++ b/docs/guides/integration-examples/test-runners.md
@@ -142,3 +142,66 @@ For AVA there is a [detailed written tutorial](https://github.com/zellwk/ava/blo
 :::note
 Note that this tutorial is pre mongodb-memory-server 7.x.
 :::
+
+## vitest
+
+For [vitest](https://vitest.dev/), create a [global setup file](https://vitest.dev/config/#globalsetup).
+
+`vitest.config.mts`:
+
+```ts
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    globalSetup: ['./globalSetup.ts'],
+  },
+});
+```
+
+`globalSetup.ts`:
+
+```ts
+import type { TestProject } from 'vitest/node';
+
+declare module 'vitest' {
+  export interface ProvidedContext {
+    MONGO_URI: string;
+  }
+}
+
+export default async function setup({ provide }: TestProject) {
+  const mongod = await MongoMemoryServer.create();
+
+  const uri = mongod.getUri();
+
+  provide('MONGO_URI', uri);
+
+  return async () => {
+    await mongod.stop();
+  };
+}
+```
+
+Then use it in your tests:
+
+`example.test.js`
+
+```ts
+import { inject, test } from 'vitest';
+import { MongoClient } from 'mongodb';
+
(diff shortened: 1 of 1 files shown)


docs/guides/integration-examples/test-runners.md
```diff
@@ -142,3 +142,66 @@ For AVA there is a [detailed written tutorial](https://github.com/zellwk/ava/blo
 :::note
 Note that this tutorial is pre mongodb-memory-server 7.x.
 :::
+
+## vitest
+
+For [vitest](https://vitest.dev/), create a [global setup file](https://vitest.dev/config/#globalsetup).
```
hasezoey, line 148: Please add a badge similar to jest and mocha:
https://github.com/typegoose/mongodb-memory-server/blob/e4992fd4f89d5025d315d05d7ca6d43358cf62be/docs/guides/integration-examples/test-runners.md?plain=1#L10

docs/guides/integration-examples/test-runners.md
```diff
@@ -142,3 +142,66 @@ For AVA there is a [detailed written tutorial](https://github.com/zellwk/ava/blo
...
+  await mongoClient.connect();
+  return () => mongoClient.disconnect();
+});
+
+test('...', () => {
+  const db = mongoClient.db('my-db');
+});
+```
+
+See also [vitest-mms](https://github.com/danielpza/vitest-mms)
```
hasezoey, line 207: Instead of just a simple `see also`, could you add a simple explanation on why this should be a `see also`, like a little description of what this package is doing / can be used?

hasezoey, line 207: Please add a note similar to jests `tests and global setup do not share a environment`:
https://github.com/typegoose/mongodb-memory-server/blob/e4992fd4f89d5025d315d05d7ca6d43358cf62be/docs/guides/integration-examples/test-runners.md?plain=1#L102-L104

re:
> Beware that the global setup is running in a different global scope, so your tests don't have access to variables defined here. However, you can pass down serializable data to tests via [provide](https://vitest.dev/config/#provide) method:
[Source: the warning](https://vitest.dev/config/#globalsetup)

docs/guides/integration-examples/test-runners.md
```diff
@@ -142,3 +142,66 @@ For AVA there is a [detailed written tutorial](https://github.com/zellwk/ava/blo
...
+  test: {
+    globalSetup: ['./globalSetup.ts'],
+  },
+});
+```
+
+`globalSetup.ts`:
+
+```ts
+import type { TestProject } from 'vitest/node';
```
hasezoey, line 165: Import for `mongodb-memory-server` missing?

Also is the type `TestProject` a global type here?

hasezoey (changes requested):
Thanks for adding a integration example for vitest. Please note that i have not used vitest yet.

> (DISCLAIMER: I'm the author of the package, I can drop it if that's an issue)

No problem, just as noted below, please update with a little description.

danielpza:
> .

Updated to address review.

PR #949, opened 2025-10-06
Update DryMongoBinary.ts

xintamosaik:
I might miss something but why is the DryMongoBinaryOption for platform fed with 'version' instead of 'platform'?

## Related Issues
Potentially Typesupport issues in IDEs that use LSP


hasezoey (approved):
Ultimately, it does not change the type from `string`, but you are correct, it should be `platform` for `platform`.

Issue #990, opened 2026-05-23
Windows: spawn EFTYPE


Sainan:
Hi, it seems ocasionally when using `mongodb-memory-server-core`, it errors with `spawn EFTYPE` (on Windows at least).

I am not quite sure how it happens that the .exe ends up being corrupted/non-executable, but it seems to happen somewhat frequently.

The best solution might be to use a checksum before deciding "oh, the .exe already exists so I don't need to download it again".

hasezoey:
I have read a bit more into it, it does not seem to be corruption related, but rather that windows expects a certain way processes are spawned and it rejects some others.
I am not yet exactly sure how or why it happens.

The next time it happen to you, could you please provide the full log of the error (including stacktrace), maybe even provide [Debug Logs](https://typegoose.github.io/mongodb-memory-server/docs/guides/enable-debug-mode/)?

PS: please dont ignore [the issue template](https://github.com/typegoose/mongodb-memory-server/blob/master/.github/ISSUE_TEMPLATE/bug.md?plain=1) and fill it out properly.

Maybe related, depending on what the error trace says: https://github.com/JuliusBrussee/cavemem/pull/24

Sainan:
It is hard to follow an issue template when I am reporting issues on behalf of users. What I personally had observed is that an .exe in my .cache/mongodb-binaries folder is missing an icon and fails to launch (seemingly truncated), but I realise now that this is unrelated to the EFTYPE error. So, I shall provide you with what I have:

# User 1

<img width="873" height="819" alt="Image" src="https://github.com/user-attachments/assets/bbcdaabc-e925-4232-99f8-07280757b84b" />

<img width="849" height="819" alt="Image" src="https://github.com/user-attachments/assets/5f6caab4-7ed5-4875-9c8b-6f2918885a36" />

<img width="772" height="819" alt="Image" src="https://github.com/user-attachments/assets/f3275c88-20df-4466-929c-e39de7e31b98" />

<img width="893" height="819" alt="Image" src="https://github.com/user-attachments/assets/ebbe0eea-d945-4446-a2fa-7c75a0725b84" />

# User 2

```
Starting the MongoMemoryServer Instance failed, enable debug log for more information. Error:
 Error: spawn EFTYPE
    at ChildProcess.spawn (node:internal/child_process:440:11)
    at spawn (node:child_process:796:9)
[...]

PR #1023, opened 2026-09-03
feat(DryMongoBinary): detect and remove corrupted cached binaries

### LINKED ISSUE #991: Option to detect corrupted binary files
As of MMS 11.1.0, we only checksum (which is the default) on download on the full archive itself, but we do no check the binaries again when trying to start them, leading to potentially corrupted binaries.
Example: let MMS download and extract it partially, kill the script and see that the file is corrupt (though that is not likely to be consistently reproducable). (re https://github.com/typegoose/mongodb-memory-server/issues/990#issuecomment-4525853881)

Implementation suggestions:
- store the checksum alongside the binaries (must be generated after extracting as checksum downloaded is for the entire archive)
- add a option to run a checksum on the binaries before each start
- add basic corrupt detection that does *not* rely on the checksum (like 0-size files or below a threshold of reasonable binary size)

Shootmewolft:
## Summary

Closes #991.

MMS only checksums the downloaded archive (`makeMD5check`), never the extracted `mongod` binary. If a process is killed mid-extraction, or a cached binary otherwise gets corrupted on disk, later runs kept trusting it from cache without any further check.

- `MongoBinaryDownload.extract()` now persists an MD5 of the extracted binary as `<binary>.md5` right after extraction (this is a checksum of the extracted binary, not the archive, since those cover different content).
- `DryMongoBinary.locateBinary()` now always rejects a cached binary below a safe minimum size (1 MiB, real `mongod` binaries are tens of MB) before trusting it, deleting the corrupted binary and its sidecar so the next run re-downloads cleanly instead of failing repeatedly on the same broken file.
- Added opt-in `MONGOMS_VALIDATE_BINARY_CHECKSUM` (default `false`) to additionally verify the persisted checksum before trusting a cached binary. Off by default because hashing a full binary on every start has a real cost; a missing sidecar (binaries cached before this change) does not fail validation, it's just skipped.

This follows the implementation suggestions from the issue.

## Test plan

- [x] `npx jest dryBinary MongoBinaryDownload` (in `packages/mongodb-memory-server-core`) — added cases for: too-small binary rejected, checksum match/mismatch when validation is enabled, missing sidecar doesn't hard-fail
- [x] `npm run lint`
- [x] `npm run build`
- [x] Documented the new option in `docs/api/config-options.md`

### DIFF
--- a/docs/api/config-options.md
+++ b/docs/api/config-options.md
@@ -294,14 +294,14 @@ Default: `true`
 ### VALIDATE_BINARY_CHECKSUM
 
 |        Environment Variable        |       PackageJson        |
-| :---------------------------------: | :-----------------------: |
+| :--------------------------------: | :----------------------: |
 | `MONGOMS_VALIDATE_BINARY_CHECKSUM` | `validateBinaryChecksum` |
 
 Option `VALIDATE_BINARY_CHECKSUM` is used to enable an md5 check of the cached (already extracted) `mongod` binary before using it, in addition to the always-on checks for obviously truncated (too small) binaries and for a missing checksum sidecar file.
 
 Default: `false`
 
-The checksum is generated once, after a binary is extracted, and stored alongside it as `<binary>.md5`. Since this sidecar is only written on a successful extraction, a cached binary without one is always treated as corrupted and re-downloaded, regardless of this option (this also applies to binaries cached by a version of this package older than this check). Enabling this option additionally hashes the full binary on every start to catch on-disk corruption after a successful extraction (e.g. a process killed mid-write); it is off by default because of that added cost.
+The checksum is generated once, after a binary is extracted, and stored alongside it as `<binary>.md5`, in the same format `md5sum --binary` writes (`CHECKSUM *FILENAME`), so it can also be verified manually with `md5sum -c <binary>.md5`. Since this sidecar is only written on a successful extraction, a cached binary without one is always treated as corrupted and re-downloaded, regardless of this option (this also applies to binaries cached by a version of this package older than this check). Enabling this option additionally hashes the full binary on every start to catch on-disk corruption after a successful extraction (e.g. a process killed mid-write); it is off by default because of that added cost.
 
 This option does not apply to a binary provided via `SYSTEM_BINARY` — system binaries are never extracted by this package, so no checksum sidecar is ever generated or checked for them.
 
--- a/docs/api/config-options.md
+++ b/docs/api/config-options.md
@@ -291,6 +291,18 @@ Option `RESUME_DOWNLOAD` is used to enable / disable resuming a download of a bi
 
 Default: `true`
 
+### VALIDATE_BINARY_CHECKSUM
+
+|         Environment Variable          |         PackageJson         |
+| :------------------------------------: | :--------------------------: |
+| `MONGOMS_VALIDATE_BINARY_CHECKSUM` | `validateBinaryChecksum` |
+
+Option `VALIDATE_BINARY_CHECKSUM` is used to enable an md5 check of the cached (already extracted) `mongod` binary before using it, in addition to the always-on check for obviously truncated (too small) binaries.
+
+Default: `false`
+
+The checksum is generated once, after a binary is extracted, and stored alongside it as `<binary>.md5`. Enabling this option adds the cost of hashing the full binary on every start, so it is off by default; use it if you suspect binaries can get corrupted on disk (e.g. a process killed mid-extraction) between runs.
+
 ## How to use them in the package.json
 
 To use the config options in the `package.json`, they need to be camelCased (and without `_`), and need to be in the property `config.mongodbMemoryServer`
--- a/docs/api/config-options.md
+++ b/docs/api/config-options.md
@@ -293,15 +293,17 @@ Default: `true`
 
 ### VALIDATE_BINARY_CHECKSUM
 
-|         Environment Variable          |         PackageJson         |
-| :------------------------------------: | :--------------------------: |
+|        Environment Variable        |       PackageJson        |
+| :---------------------------------: | :-----------------------: |
 | `MONGOMS_VALIDATE_BINARY_CHECKSUM` | `validateBinaryChecksum` |
 
-Option `VALIDATE_BINARY_CHECKSUM` is used to enable an md5 check of the cached (already extracted) `mongod` binary before using it, in addition to the always-on check for obviously truncated (too small) binaries.
+Option `VALIDATE_BINARY_CHECKSUM` is used to enable an md5 check of the cached (already extracted) `mongod` binary before using it, in addition to the always-on checks for obviously truncated (too small) binaries and for a missing checksum sidecar file.
 
 Default: `false`
 
-The checksum is generated once, after a binary is extracted, and stored alongside it as `<binary>.md5`. Enabling this option adds the cost of hashing the full binary on every start, so it is off by default; use it if you suspect binaries can get corrupted on disk (e.g. a process killed mid-extraction) between runs.
+The checksum is generated once, after a binary is extracted, and stored alongside it as `<binary>.md5`. Since this sidecar is only written on a successful extraction, a cached binary without one is always treated as corrupted and re-downloaded, regardless of this option (this also applies to binaries cached by a version of this package older than this check). Enabling this option additionally hashes the full binary on every start to catch on-disk corruption after a successful extraction (e.g. a process killed mid-write); it is off by default because of that added cost.
+
+This option does not apply to a binary provided via `SYSTEM_BINARY` — system binaries are never extracted by this package, so no checksum sidecar is ever generated or checked for them.
 
 ## How to use them in the package.json
 
--- a/packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
+++ b/packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
@@ -131,46 +131,69 @@ export class DryMongoBinary {
     }
 
     const foundBinaryPath = returnValue[1];
-    const binaryStat = await statPath(foundBinaryPath);
 
-    if (!isNullOrUndefined(binaryStat) && binaryStat.size < MIN_BINARY_SIZE_BYTES) {
-      log(
-        `locateBinary: binary at "${foundBinaryPath}" is only "${binaryStat.size}" bytes (expected at least "${MIN_BINARY_SIZE_BYTES}"), treating as corrupted`
-      );
+    if (await this.isBinaryCorrupted(foundBinaryPath)) {
       await this.removeCorruptedBinary(foundBinaryPath);
 
       return undefined;
-    }
-
-    // opt-in, default off: hashing the full binary on every start has a real cost
-    if (envToBool(resolveConfig(ResolveConfigVariables.VALIDATE_BINARY_CHECKSUM))) {
-      const checksumFile = `${foundBinaryPath}.md5`;
-
-      if (await pathExists(checksumFile)) {
-        const [expectedChecksum, actualChecksum] = await Promise.all([
-          fspromises.readFile(checksumFile, 'utf-8'),
-          md5FromFile(foundBinaryPath),
-        ]);
-
-        if (expectedChecksum.trim() !== actualChecksum) {
-          log(
-            `locateBinary: checksum mismatch for binary at "${foundBinaryPath}" (expected: "${expectedChecksum.trim()}", actual: "${actualChecksum}"), treating as corrupted`
-          );
-          await this.removeCorruptedBinary(foundBinaryPath);
-
-          return undefined;
-        }
-      } else {
-        log(
-          `locateBinary: "VALIDATE_BINARY_CHECKSUM" is enabled, but no checksum file exists for "${foundBinaryPath}", skipping validation`
-        );
-      }
     }
 
     log(`locateBinary: found binary at "${foundBinaryPath}"`);
     this.binaryCache.set(opts.version, foundBinaryPath);
 
     return foundBinaryPath;
+  }
+
+  /**
+   * Check whether the binary at "binaryPath" is corrupted:
+   * - it cannot be stat'd
+   * - it is below the expected minimum size (truncated download/extraction)
+   * - its checksum sidecar (`<binary>.md5`) is missing (it is always written on successful
+   *   extraction, so its absence means an incomplete/pre-existing extraction)
+   * - (opt-in via "VALIDATE_BINARY_CHECKSUM") its content does not match the checksum sidecar
+   */
+  private static async isBinaryCorrupted(binaryPath: string): Promise<boolean> {
+    const binaryStat = await statPath(binaryPath);
+
+    if (isNullOrUndefined(binaryStat)) {
+      log(`isBinaryCorrupted: could not stat binary at "${binaryPath}", treating as corrupted`);
+
+      return true;
+    }
+
+    if (binaryStat.size < MIN_BINARY_SIZE_BYTES) {
+      log(
+        `isBinaryCorrupted: binary at "${binaryPath}" is only "${binaryStat.size}" bytes (expected at least "${MIN_BINARY_SIZE_BYTES}"), treating as corrupted`
+      );
+
+      return true;
+    }
+
+    const checksumFile = `${binaryPath}.md5`;
+
+    if (!(await pathExists(checksumFile))) {
+      log(`isBinaryCorrupted: no checksum file exists for "${binaryPath}", treating as corrupted`);
+
+      return true;
+    }
+
+    // opt-in, default off: hashing the full binary on every start has a real cost
+    if (envToBool(resolveConfig(ResolveConfigVariables.VALIDATE_BINARY_CHECKSUM))) {
+      const [expectedChecksum, actualChecksum] = await Promise.all([
+        fspromises.readFile(checksumFile, 'utf-8'),
+        md5FromFile(binaryPath),
+      ]);
+
+      if (expectedChecksum.trim() !== actualChecksum) {
+        log(
+          `isBinaryCorrupted: checksum mismatch for binary at "${binaryPath}" (expected: "${expectedChecksum.trim()}", actual: "${actualChecksum}"), treating as corrupted`
+        );
+
+        return true;
+      }
+    }
+
+    return false;
   }
 
   /**
--- a/packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
+++ b/packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
@@ -1,26 +1,31 @@
 import debug from 'debug';
+import findCacheDir from 'find-cache-dir';
+import { promises as fspromises } from 'fs';
+import { arch, homedir, platform } from 'os';
+import * as path from 'path';
+import { BinaryNotFoundError, NoRegexMatchError, ParseArchiveRegexError } from './errors';
+import { AnyOS, getOS, isLinuxOS, OtherOS } from './getos';
+import { MongoBinaryDownloadUrl } from './MongoBinaryDownloadUrl';
 import {
   DEFAULT_VERSION,
   envToBool,
   packageJsonPath,
-  resolveConfig,
   ResolveConfigVariables,
+  resolveConfig,
 } from './resolveConfig';
 import {
   assertion,
   checkBinaryPermissions,
   isNullOrUndefined,
   lockfilePath,
+  md5FromFile,
   pathExists,
+  statPath,
 } from './utils';
-import * as path from 'path';
-import { arch, homedir, platform } from 'os';
-import findCacheDir from 'find-cache-dir';
-import { getOS, AnyOS, isLinuxOS, OtherOS } from './getos';
-import { BinaryNotFoundError, NoRegexMatchError, ParseArchiveRegexError } from './errors';
-import { MongoBinaryDownloadUrl } from './MongoBinaryDownloadUrl';
 
 const log = debug('MongoMS:DryMongoBinary');
+
+const MIN_BINARY_SIZE_BYTES = 1024 * 1024;
 
 /** Interface for general options required to pass-around (all optional) */
 export interface BaseDryMongoBinaryOptions {
@@ -124,10 +129,55 @@ export class DryMongoBinary {
       return undefined;
     }
 
-    log(`locateBinary: found binary at "${returnValue[1]}"`);
-    this.binaryCache.set(opts.version, returnValue[1]);
+    const foundBinaryPath = returnValue[1];
+    const binaryStat = await statPath(foundBinaryPath);
 
-    return returnValue[1];
+    if (!isNullOrUndefined(binaryStat) && binaryStat.size < MIN_BINARY_SIZE_BYTES) {
+      log(
+        `locateBinary: binary at "${foundBinaryPath}" is only "${binaryStat.size}" bytes (expected at least "${MIN_BINARY_SIZE_BYTES}"), treating as corrupted`
+      );
+      await this.removeCorruptedBinary(foundBinaryPath);
+
+      return undefined;
+    }
+
+    // opt-in, default off: hashing the full binary on every start has a real cost
+    if (envToBool(resolveConfig(ResolveConfigVariables.VALIDATE_BINARY_CHECKSUM))) {
+      const checksumFile = `${foundBinaryPath}.md5`;
+
+      if (await pathExists(checksumFile)) {
+        const [expectedChecksum, actualChecksum] = await Promise.all([
+          fspromises.readFile(checksumFile, 'utf-8'),
+          md5FromFile(foundBinaryPath),
+        ]);
+
+        if (expectedChecksum.trim() !== actualChecksum) {
+          log(
+            `locateBinary: checksum mismatch for binary at "${foundBinaryPath}" (expected: "${expectedChecksum.trim()}", actual: "${actualChecksum}"), treating as corrupted`
+          );
+          await this.removeCorruptedBinary(foundBinaryPath);
+
+          return undefined;
+        }
+      } else {
+        log(
+          `locateBinary: "VALIDATE_BINARY_CHECKSUM" is enabled, but no checksum file exists for "${foundBinaryPath}", skipping validation`
+        );
+      }
+    }
+
+    log(`locateBinary: found binary at "${foundBinaryPath}"`);
+    this.binaryCache.set(opts.version, foundBinaryPath);
+
+    return foundBinaryPath;
+  }
+
+  /**
+   * Remove a corrupted binary and its checksum sidecar
+   */
+  private static async removeCorruptedBinary(binaryPath: string): Promise<void> {
+    await fspromises.rm(binaryPath, { force: true });
+    await fspromises.rm(`${binaryPath}.md5`, { force: true });
   }
 
   /**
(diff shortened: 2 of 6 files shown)


docs/api/config-options.md
```diff
@@ -291,6 +291,18 @@ Option `RESUME_DOWNLOAD` is used to enable / disable resuming a download of a bi
 
 Default: `true`
 
+### VALIDATE_BINARY_CHECKSUM
+
+|         Environment Variable          |         PackageJson         |
+| :------------------------------------: | :--------------------------: |
+| `MONGOMS_VALIDATE_BINARY_CHECKSUM` | `validateBinaryChecksum` |
```
hasezoey:
```suggestion
|        Environment Variable        |       PackageJson        |
| :--------------------------------: | :----------------------: |
| `MONGOMS_VALIDATE_BINARY_CHECKSUM` | `validateBinaryChecksum` |
```

packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
```diff
@@ -1,27 +1,32 @@
...
-import * as path from 'path';
-import { arch, homedir, platform } from 'os';
-import findCacheDir from 'find-cache-dir';
-import { getOS, AnyOS, isLinuxOS, OtherOS } from './getos';
-import { BinaryNotFoundError, NoRegexMatchError, ParseArchiveRegexError } from './errors';
-import { MongoBinaryDownloadUrl } from './MongoBinaryDownloadUrl';
 
 const log = debug('MongoMS:DryMongoBinary');
 
+const MIN_BINARY_SIZE_BYTES = 1024 * 1024;
```
hasezoey, line 29: Some documentation would be good, like what is this minimum for? (it is obvious in this PR, but on a glance it might not be)

packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
```diff
@@ -124,10 +129,55 @@ export class DryMongoBinary {
       return undefined;
     }
 
-    log(`locateBinary: found binary at "${returnValue[1]}"`);
-    this.binaryCache.set(opts.version, returnValue[1]);
+    const foundBinaryPath = returnValue[1];
+    const binaryStat = await statPath(foundBinaryPath);
 
-    return returnValue[1];
+    if (!isNullOrUndefined(binaryStat) && binaryStat.size < MIN_BINARY_SIZE_BYTES) {
```
hasezoey: Shouldnt this also fail early if no stat can be made?

packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
```diff
@@ -124,10 +129,55 @@ export class DryMongoBinary {
...
+        ]);
+
+        if (expectedChecksum.trim() !== actualChecksum) {
+          log(
+            `locateBinary: checksum mismatch for binary at "${foundBinaryPath}" (expected: "${expectedChecksum.trim()}", actual: "${actualChecksum}"), treating as corrupted`
+          );
+          await this.removeCorruptedBinary(foundBinaryPath);
+
+          return undefined;
+        }
```
hasezoey: Consider putting this code into a separate function

packages/mongodb-memory-server-core/src/util/__tests__/dryBinary.test.ts
```diff
@@ -585,6 +586,76 @@ describe('DryBinary', () => {
       expect(binary.DryMongoBinary.binaryCache.has).toHaveBeenCalledTimes(1);
       expect(binary.DryMongoBinary.generateDownloadPath).toHaveBeenCalled();
     });
+
+    it('should return "undefined" and remove the binary if it is below the minimum size threshold', async () => {
+      jest
+        .spyOn(binary.DryMongoBinary, 'generateDownloadPath')
+        .mockResolvedValue([true, '/tmp/1.1.1']);
+      jest.spyOn(utils, 'pathExists').mockImplementation((path) => {
+        return Promise.resolve(path === '/tmp/1.1.1.lock' ? false : false);
```
hasezoey: This only returns `false`, why the condition?

docs/api/config-options.md
```diff
@@ -291,6 +291,18 @@ Option `RESUME_DOWNLOAD` is used to enable / disable resuming a download of a bi
 
 Default: `true`
 
+### VALIDATE_BINARY_CHECKSUM
```
hasezoey, line 294: Just to be clear, this option should note that it does not check `SYSTEM_BINARY` checksums (as they likely dont exist)

packages/mongodb-memory-server-core/src/util/__tests__/dryBinary.test.ts
hasezoey, line 1: I dont quite like that there is so much file mocking.
Consider using tempdirs instead so that if something in the runtime code actually changes (ex. adding more `stat` on different files), this does not silently blow up.

hasezoey (changes requested):
Thanks for making this feature, could you:
- change base branch to be `beta`
- split the MIN_SIZE check into a different commit

As a additional check, what do you think of just assuming the binary is corrupt if no checksum file exists (even with `VALIDATE_BINARY_CHECKSUM` disabled)? This would be possible as the file only exists if the extraction was successful.
Though that gives the issue of existing installations, the options i can think for for them would be:
1. the safest option: force a redownload of the binaries
2. the fastest option: if the binary file was created before this PR's date and passes the other checks (ex. MIN_SIZE), assume it is non-corrupt and generate a checksum for it

WDYT?

Shootmewolft:
Thanks for the detailed review! Addressed everything:

- Retargeted the PR to `beta`.
- Split the `MIN_SIZE` check into its own commit (906b7932), separate from the checksum work (d994d467).
- Added documentation for `MIN_BINARY_SIZE_BYTES`.
- Extracted the corruption logic into a dedicated `isBinaryCorrupted` method.
- Added a fail-early path for when the binary can't be `stat`'d at all.
- On the "assume corrupt if no checksum file exists" question: went with the safe option. A missing `.md5` now always forces a re-download, regardless of `VALIDATE_BINARY_CHECKSUM` — this covers binaries cached by older versions of this package too, since none of them ever wrote a checksum sidecar. Docs updated to reflect this.
- Fixed the no-op ternary in the min-size test (it always returned `false` either way).
- Replaced the `statPath`/`pathExists`/`md5FromFile` mocking in the corruption-detection tests with real tempdirs and real files (e98308fa), so a future added `stat`/path check can't silently pass through a stale mock. I scoped this to the tests this PR introduces (`locateBinary`'s corruption-detection suite) rather than the whole file, to avoid touching unrelated pre-existing tests — happy to extend it further if you'd rather have the whole file converted.

Let me know if the re-download-on-missing-checksum behavior is what you had in mind, or if you'd prefer the "generate checksum on the fly" approach for existing installs instead.

packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
```diff
@@ -25,6 +25,7 @@ import {
 
 const log = debug('MongoMS:DryMongoBinary');
 
+// a real "mongod" binary is always well above this size; anything smaller indicates a truncated download/extraction
```
hasezoey:
```suggestion
/** a real "mongod" binary is always well above this size; anything smaller indicates a truncated download/extraction */
```

Otherwise it is not treated as a JSDoc comment.

packages/mongodb-memory-server-core/src/util/__tests__/dryBinary.test.ts
```diff
@@ -587,74 +591,86 @@ describe('DryBinary', () => {
...
+
+      it('should return the binary path when checksum validation is enabled and the checksum matches', async () => {
+        process.env[envName(ResolveConfigVariables.VALIDATE_BINARY_CHECKSUM)] = 'true';
+        await fspromises.writeFile(binaryPath, Buffer.alloc(2 * 1024 * 1024, 'a'));
+        await fspromises.writeFile(`${binaryPath}.md5`, await utils.md5FromFile(binaryPath));
+
+        const returnValue = await binary.DryMongoBinary.locateBinary({ version: '1.1.1' });
+        expect(returnValue).toEqual(binaryPath);
+        expect(binary.DryMongoBinary.binaryCache.size).toBe(1);
+      });
```
hasezoey, line 671: This should test that `md5FromFile` has been called.

packages/mongodb-memory-server-core/src/util/__tests__/dryBinary.test.ts
```diff
@@ -587,74 +591,86 @@ describe('DryBinary', () => {
...
+        process.env[envName(ResolveConfigVariables.VALIDATE_BINARY_CHECKSUM)] = 'true';
+        await fspromises.writeFile(binaryPath, Buffer.alloc(2 * 1024 * 1024, 'a'));
+        await fspromises.writeFile(`${binaryPath}.md5`, 'not-the-real-checksum');
+
+        const returnValue = await binary.DryMongoBinary.locateBinary({ version: '1.1.1' });
+        expect(returnValue).toBeUndefined();
+        expect(binary.DryMongoBinary.binaryCache.size).toBe(0);
+        expect(await utils.pathExists(binaryPath)).toBe(false);
+        expect(await utils.pathExists(`${binaryPath}.md5`)).toBe(false);
+      });
```
hasezoey, line 685: This should also check that the `md5FromFile` function has been called

packages/mongodb-memory-server-core/src/util/__tests__/dryBinary.test.ts
```diff
@@ -587,74 +591,86 @@ describe('DryBinary', () => {
...
-      jest
-        .spyOn(binary.DryMongoBinary, 'generateDownloadPath')
-        .mockResolvedValue([true, '/tmp/1.1.1']);
-      jest.spyOn(utils, 'pathExists').mockImplementation((path) => {
-        return Promise.resolve(path === '/tmp/1.1.1.lock' ? false : false);
+    // These tests use a real tempdir and real files instead of mocking "statPath"/"pathExists"/
+    // "md5FromFile", so that a future change to "isBinaryCorrupted" (ex. adding another "stat"
+    // or a check on a different path) cannot silently pass because a mock kept returning a
+    // stale canned value.
+    describe('corruption detection', () => {
```
hasezoey, line 598: There likely should also be a test case for checksum verification being disabled and still being successful, maybe even a test where the checksum does not match but verification is disabled too.

hasezoey (changes requested):
> Split the MIN_SIZE check into its own commit (https://github.com/typegoose/mongodb-memory-server/commit/906b7932947c890c6e443f3770a3b348d29f4ba7), separate from the checksum work (https://github.com/typegoose/mongodb-memory-server/commit/d994d4676a3d94e80d9192b7ba8ae48a9b766ebe).

Does not look that way to me (906b7932947c890c6e443f3770a3b348d29f4ba7 only adds the documentation, whereas 3d5bf0bc0823363edc3671624389d7ba259af869 still has the MIN_SIZE check).
Though that is not a big matter and does not need changing anymore.

> I scoped this to the tests this PR introduces (locateBinary's corruption-detection suite) rather than the whole file, to avoid touching unrelated pre-existing tests — happy to extend it further if you'd rather have the whole file converted.

No need to touch non-PR related tests.

---

FYI, you can close threads if you think you fully resolved them.

Currently this outputs the md5sum raw, should we maybe output it in the `cksum` (`md5sum` / `shaXsum`) format of `CHECKSUM  FILENAME` (note the double space inbetween) or `CHECKSUM *FILENAME`(note the asterist for binary mode)?
(Note from my quick test of a random mongodb binary via `md5sum` cli, both "text mode" and "binary mode" result in the same checksum for me, i am not *quite* sure what mode we are using here)

Shootmewolft:
Thanks for the follow-up — you're right about the commit split, I misread the ask and only layered new commits on top instead of actually separating `MIN_SIZE` out of 3d5bf0b. Noted for next time.

Addressed in a2e53fb:

- **Checksum file format**: now written as `CHECKSUM *FILENAME` (`md5sum --binary` format), so it can be verified manually with `md5sum -c`. On the mode question: text vs binary mode only differ where the C runtime does newline translation, which Node's `fs` never does — so the checksum is identical either way, and `*` is just the semantically correct marker for a binary file. Reading only takes the first whitespace-separated token, so sidecars written before this change (bare hash, no filename) still validate — there's a test covering that.
- Made the `MIN_BINARY_SIZE_BYTES` comment a proper JSDoc block.
- Both checksum tests now assert `md5FromFile` was actually called.
- Added the missing cases: `VALIDATE_BINARY_CHECKSUM` disabled with a matching checksum, and disabled with a *mismatching* checksum — both succeed without hashing.
- While re-checking your table suggestion, I noticed the separator row I had applied was 2 characters wider than the other rows, so it wasn't quite your suggested block. Fixed.
- Documented the sidecar format in the config docs.

I did not run Prettier over `docs/api/config-options.md`: it has pre-existing formatting diffs unrelated to this PR (and markdown isn't covered by `lint-staged` or CI here), so reformatting it would have added noise to the diff.

Closing the threads I believe are fully resolved — reopen any I got wrong.

packages/mongodb-memory-server-core/src/util/__tests__/dryBinary.test.ts
```diff
@@ -653,24 +661,61 @@ describe('DryBinary', () => {
...
         const returnValue = await binary.DryMongoBinary.locateBinary({ version: '1.1.1' });
         expect(returnValue).toBeUndefined();
         expect(binary.DryMongoBinary.binaryCache.size).toBe(0);
+        expect(md5Spy).toHaveBeenCalledWith(binaryPath);
         expect(await utils.pathExists(binaryPath)).toBe(false);
         expect(await utils.pathExists(`${binaryPath}.md5`)).toBe(false);
       });
+
+      it('should return the binary path when checksum validation is disabled and an existing checksum matches', async () => {
+        await fspromises.writeFile(binaryPath, Buffer.alloc(2 * 1024 * 1024, 'a'));
```
hasezoey, line 688: As this documentation explicitly notes "disabled validation", this should explicitly set the check to be disabled instead of relying on the default (which may get changed in the future).
Add:
```ts
process.env[envName(ResolveConfigVariables.VALIDATE_BINARY_CHECKSUM)] = 'false';
```

And any other test that may do the same explicit testing.

packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
```diff
@@ -179,14 +179,17 @@ export class DryMongoBinary {
 
     // opt-in, default off: hashing the full binary on every start has a real cost
     if (envToBool(resolveConfig(ResolveConfigVariables.VALIDATE_BINARY_CHECKSUM))) {
-      const [expectedChecksum, actualChecksum] = await Promise.all([
+      const [checksumFileContent, actualChecksum] = await Promise.all([
         fspromises.readFile(checksumFile, 'utf-8'),
         md5FromFile(binaryPath),
       ]);
+      // the sidecar is written as "CHECKSUM *FILENAME" (md5sum's binary-mode format), but a sidecar
+      // written before that format was introduced only has the raw checksum, so only ever read the first token
```
hasezoey, line 187:
```suggestion
      // The sidecar is written as "CHECKSUM *FILENAME" (md5sum's binary-mode format).
      // As we know the file we want to check, we can safely ignore the filename.
```

As this PR introduces the checksum file at all, i dont think we need to state that this supports a "legacy" format.
Another suggestion would be to also just check the filename to also match.

hasezoey (changes requested):
Looks good, i think this is the last round of changes necessary.

Just for disclosure, is this (at least the github comments) AI generated?

Shootmewolft:
Yes — the GitHub comments are AI-assisted, and so are parts of the code: inline comments, tests, and doc wording. I direct the work, review every diff, and run tests and lint locally before pushing. I should have disclosed that upfront instead of waiting to be asked.

Worth flagging one case where my process failed, since you caught it: the "legacy bare-hash sidecar" backward-compatibility claim in my previous comment was wrong. I checked `beta` after reading your comment — there is no binary sidecar there at all, only the archive `makeMD5check`. So no legacy format ever existed in a released version, and the compatibility test I added rests on a false premise. Your suggestion is right; I'll apply it and drop that test.

If you'd rather I handle this differently in this repo, tell me and I'll adjust.

packages/mongodb-memory-server-core/src/util/__tests__/dryBinary.test.ts
Shootmewolft: Done — the three tests that state "disabled" in their name now set `VALIDATE_BINARY_CHECKSUM` to `false` explicitly. I left the min-size and unstat-able tests alone since neither claims anything about the validation state.

packages/mongodb-memory-server-core/src/util/DryMongoBinary.ts
Shootmewolft: Applied your suggestion. You are right that there is no legacy format to support — I checked `beta` and the binary sidecar does not exist there at all, only the archive `makeMD5check`, so nothing released ever wrote one. I also dropped the legacy-format test, since it was testing an input that cannot occur.

I kept ignoring the filename rather than validating it, since that is what your suggestion describes and the file under verification is already known. Happy to switch to checking it if you prefer the stricter version.

hasezoey (changes requested):
Overall, now looks good to me.
I noticed that the checksum gets generated *after* extraction on final binary, which would for example not catch #990, where the extraction truncated the file (not writing the last chunk). Could the md5 generation be piped directly into the extraction + write pipeline of both tar and zip extractions?
Sorry that i noticed this rather late.

Shootmewolft:
You're right, and it goes deeper than the timing of the hash: neither extraction path waits for the write to finish.

In `extractTarGz` the entry handler pipes into the write stream but calls `next()` on the *source* stream's `end`, and the outer promise resolves on the tar stream's `finish` — that is when parsing completes, not when the bytes reach disk:

```js
stream.pipe(createWriteStream(extractPath, { mode: 0o775 }));
stream.on('end', () => next());
// ...
.pipe(extract).on('finish', res)
```

`extractZip` has the same shape (`r.on('end', () => zipfile.readEntry())` plus `zipfile.on('end', () => resolve())`). Neither write stream has an `error` handler either, so a failed write would surface as an unhandled stream error while the extraction still reports success.

So `extract()` can reach `pathExists` and `md5FromFile` while the write is still in flight. That is both a plausible mechanism for #990 and the reason the sidecar can end up certifying a truncated binary.

One consequence worth flagging: piping the md5 into the extraction alone would not close this. With `VALIDATE_BINARY_CHECKSUM` off by default, a truncated binary with a matching sidecar still passes `isBinaryCorrupted` — the sidecar exists, and if only the last chunk is missing the file is still above `MIN_SIZE`. The inline hash also needs the write to be awaited first: comparing it against the on-disk file while the write is in flight would fail intermittently on perfectly good downloads, trading a rare corruption for a rare false failure.

Plan, as two separate commits so they can be split off if you prefer:

1. Await `finish`/`close` on the write streams and attach `error` handlers in both extraction paths, so an incomplete write fails the extraction instead of being cached.
2. Pipe the md5 through the extraction as you suggested, and verify it against the written file before the sidecar is created.

Happy to move either or both into a separate PR targeting #990 if you would rather keep this one scoped — just say which.

Shootmewolft:
Correction to my previous comment, and the first commit (b8987701).

I could not reproduce the truncation I suggested was a plausible mechanism for #990. I wrote tests that extract a 4-8 MB entry and read the file size the moment the extraction resolves (with `statSync`, so nothing is awaited in between that would let the queued writes drain) — they pass with and without the fix. The likely reason is that the source side (gunzip / the inflate in yauzl) is slower than the write, so by the time parsing finishes the write has already drained. The window is real but I have no way to force it: the write stream is created inside the extraction methods, so there is no injection point, and `fs.createWriteStream` cannot be spied on (`TypeError: Cannot redefine property`). I dropped those tests rather than keep ones that pass either way.

What is deterministic and reproducible is what happens when the write actually fails. Pointing the extraction target at a directory (EISDIR on open), on the current code:

- `extractZip` never settles — the extraction hangs forever. When the read stream is destroyed, `r` never emits `end`, so `readEntry()` is never called again, so `zipfile` never emits `end`, so the promise is left pending.
- Both paths also emit the EISDIR as an unhandled stream error.

So the defect I can actually demonstrate is not a corrupt binary, it is a failed write turning into a hang plus an out-of-band error. b8987701 pipes through `stream/promises`, awaits the write after the archive is parsed, and rejects the extraction when the write fails. Worth noting that awaiting alone was not enough: my first attempt removed the unhandled error but still hung, because nothing connected the write failure to the outer promise's `reject`.

Two regression tests cover it (`should reject when the binary from a {zip,tar.gz} archive cannot be written`). I checked that both fail on the pre-fix code rather than assuming it.

The inline md5 you asked for is next, as the second commit. Given the above, I would frame it as making the sidecar describe what the archive contained rather than whatever ended up on disk, instead of claiming it fixes #990 — I do not have evidence for that claim. If you know the conditions under which the truncation in #990 shows up, I would rather reproduce it first than guess.

Shootmewolft:
Second commit is up (867ea8f4): the md5 is now taken as the entry passes through the extraction pipeline, via a passthrough that feeds the hash, instead of reading the file back afterwards. Before the sidecar is written, that checksum is compared against the file that was actually written, and a mismatch throws instead of being cached.

The sidecar contents are unchanged — this only changes which bytes the checksum is taken from, plus the comparison. Note that the comparison needed the first commit to be in place: while the write was not awaited, comparing against the on-disk file could have failed on perfectly good downloads.

`extractTarGz` and `extractZip` now return the checksum of the extracted entry (`Promise<string | undefined>` instead of `Promise<void>`); `undefined` when no entry matched the filter. They are public, so tell me if you would rather keep their signature and surface it some other way.

Tests: a new case asserts the extraction throws when the written binary does not match what came out of the archive, and that no sidecar is left behind in that case. I checked it fails on the pre-commit code. 302 tests pass locally (everything under `src/util` except `MongoInstance`, which fails here for an unrelated missing `libcrypto.so.1.1`).

On #990: I would rather not claim these two commits close it, since I still have not reproduced the truncation. What they do establish is that a failed write now fails loudly instead of hanging, and that the sidecar describes what the archive contained rather than whatever landed on disk. If you can share the conditions where you have seen the truncated extraction, I will try to reproduce it against this branch.

Shootmewolft:
Merged `beta` in (f18f9191) — the branch had gone stale against the beta.3 release and was conflicting. I merged rather than rebased so the commits you already reviewed keep their hashes.

The conflicts were only in the extraction code and its tests, and resolving them changed what this branch contributes:

- **`extractZip`**: 3fb04f4b already replaced it with yauzl's promise API, which awaits the write stream's `finish` and handles its errors. That is the same defect the zip half of b8987701 was fixing, so I took your implementation as-is and dropped mine. Yours is the better version — the `for await` over `eachEntry()` reads much more clearly than what I had.
- **`extractTarGz`**: untouched upstream, and it still resolves when the archive finishes parsing rather than when the write completes, with no error handler on the write. So that fix stays.
- **The extraction-time md5** is re-applied on top of your zip implementation, following its style (explicit `pipe` plus `once` handlers) instead of reintroducing `stream/promises` there.

One thing I could not fix without rewriting history: b8987701's message says "Both extraction paths", which is no longer true after the merge — only its tar half survives. I spelled that out in the merge commit body rather than leave the message misleading on its own.

Verification on the merged tree, not on my earlier local rebase: 302 tests pass under `src/util` (all but `MongoInstance`, which fails here for an unrelated missing `libcrypto.so.1.1`), typecheck and lint are clean. I also re-reverted the tar await locally to confirm `should reject when the binary from a tar.gz archive cannot be written` still fails without it. The zip write-failure test now passes against your implementation, so it stands as a guard on that code rather than on mine — happy to drop it if you would rather not carry it.

Note for anyone else pulling this branch: `npm ci` is needed after the merge, since beta moved yauzl to 3.4.0 and `openPromise` does not typecheck against the older types.

packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
```diff
@@ -286,6 +311,15 @@ export class MongoBinaryDownload {
 
     // checksum of the extracted binary, not the archive (that is checked in "makeMD5check")
     const binaryChecksum = await md5FromFile(mongodbFullPath);
+
+    // "extractedChecksum" was taken from the archive as the binary was extracted, so a difference
+    // means what was written is not what the archive contained
+    if (!isNullOrUndefined(extractedChecksum) && extractedChecksum !== binaryChecksum) {
+      throw new GenericMMSError(
+        `MongoBinaryDownload: the extracted binary at "${mongodbFullPath}" does not match what was extracted from "${mongoDBArchive}" (expected: "${extractedChecksum}", actual: "${binaryChecksum}")`
```
hasezoey:
```suggestion
        `MongoBinaryDownload: the extracted binary at "${mongodbFullPath}" does not match what was extracted from "${mongoDBArchive}"! This likely means there has been a bug or your filesystem is bad, please report this! (expected: "${extractedChecksum}", actual: "${binaryChecksum}")`
```

packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
```diff
@@ -305,23 +339,34 @@ export class MongoBinaryDownload {
     mongoDBArchive: string,
     extractPath: string,
     filter: (file: string) => boolean
-  ): Promise<void> {
+  ): Promise<string | undefined> {
```
hasezoey: The TSDoc should mention *what* the returned value is.

packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownload.test.ts
```diff
@@ -434,11 +435,98 @@ describe('MongoBinaryDownload', () => {
...
+      jest.spyOn(utils, 'md5FromFile').mockResolvedValue('0123456789abcdef0123456789abcdef');
+
+      await expect(mbd.extract(zipPath)).rejects.toThrow(/does not match what was extracted/);
+      // the sidecar must not be created for a binary that failed verification
+      expect(await utils.pathExists(`${outPath}.md5`)).toStrictEqual(false);
+    });
+
+    // Regression tests for write failures during extraction. Previously the write stream was piped
+    // without anything awaiting or handling it, so a failed write surfaced as an unhandled stream
+    // error and left the extraction pending forever instead of failing it. See #990.
```
hasezoey: I am pretty sure this has nothing todo with #990. That issue, from what i remember and a quick skim, is about truncated output, not about a hanging extraction.

hasezoey (approved):
LGTM. Frankly your last comments were a lot more text than the actual changes.
Some minor things, which i will quickly fix myself.

packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
hasezoey: Fixed with 9f8a39582ded857265b9d76b946c0b8ef6cd18f3.

packages/mongodb-memory-server-core/src/util/MongoBinaryDownload.ts
hasezoey: Fixed with 9f8a39582ded857265b9d76b946c0b8ef6cd18f3.

packages/mongodb-memory-server-core/src/util/__tests__/MongoBinaryDownload.test.ts
hasezoey: Removed the `See #990` note as i dont think that was relevant.

